PR #2, opened 2021-03-23
add DBSecurityGroup resource

jaypipes:
Adds basic create and delete operations for DBSecurityGroup resource in
the RDS API. This resource has more attributes than name and
description, such as EC2 security group and IPRange fields, however
there are separate Allow and Revoke Operation API calls for these fields
and will need to be handled in custom code.

aws-controllers-k8s/community#237


A-Hilaly (approved):
:+1:

A-Hilaly:
one tiny nit: we can re-generate using the `code-generator` latest commit.

Also i'm thinking that maybe we should leave a "footprint" in the generated code representing the used `ack-generate` version/commit. I believe that it will make reviewing similar PRs easier. WDYT @jaypipes ?

jaypipes:
> one tiny nit: we can re-generate using the `code-generator` latest commit.
> 
> Also i'm thinking that maybe we should leave a "footprint" in the generated code representing the used `ack-generate` version/commit. I believe that it will make reviewing similar PRs easier. WDYT @jaypipes ?

I re-gen'd the RDS controller with ACK code-generator v0.0.6 (includes the docstring generation). I then rebased this commit onto that. Please re-review and merge if OK (you are now an RDS controller maintainer, btw)

A-Hilaly (approved):

scandeezy:
I've been looking at this because I'm intending to setup a security group for accessing the RDS instance.  Without this code, I see it sets to the default VPC security Group.

I can manually/cloudformation/terraform the security group and declare this as part of the dbinstance, but I'm hoping to control access to each db separately.  Is there a new pattern I should be made aware of?

PR #10, opened 2021-05-25
add DBInstance resource

jaypipes:
Adds support for the DBInstance custom resource. There are a number of
callouts for this CRD:

1) The DBSecurityGroups field of the CreateDBInstance API Input shape
   has been renamed to DBSecurityGroupNames in order to bypass an issue
   where a same-named field in the CreateDBInstance API Output shape was of
   a different data type. The DBSecurityGroups field in the
   CreateDBInstance API Output shape now appears as
   `Status.DBSecurityGroups` and has a data type that shows the
   DBSecurityGroup name and "status".

2) The MasterUserPassword field has been replaced with
   SecretKeyReference.

3) Hooks for sdk_update_pre_build_request and
   sdk_delete_pre_build_request have been added that check the
   DBInstanceStatus field. The valid values for this field had to be
   derived from the RDS control plane source code since there is no enum
   in the public API model.

4) The SkipFinalSnapshot field is always set to true in the
   DeleteDBInstance operation. This is not ideal but will suffice until
   we develop a method to generate a DBSnapshot name automatically.

Issue aws-controllers-k8s/community#752


pkg/resource/db_instance/hooks.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+var (
+	requeueWaitWhileDeleting = ackrequeue.NeededAfter(
+		errors.New("DB instance in 'deleting' state, cannot be modified or deleted."),
+		ackrequeue.DefaultRequeueAfterDuration,
+	)
+)
+
+// requeueWaitWhileCreating returns a `ackrequeue.RequeueNeededAfter` struct
```
vijtrip2, line 95: nit: method names mismatch

vijtrip2 (approved):

pkg/resource/db_instance/hooks.go
jaypipes: copy/pasta :) thanks for the heads up!

generator.yaml
```diff
@@ -16,7 +16,41 @@ ignore:
     - EventSubscription
     - GlobalCluster
     - OptionGroup
+  field_paths:
```
A-Hilaly, line 19: nit: unused configuration

test/e2e/tests/test_db_instance.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture(scope="module")
+def master_user_pass_secret():
+    ns = "default"
+    name = "dbsecrets"
+    key = "master_user_password"
+    secret_val = "secretpass123456"
+    k8s.create_opaque_secret(ns, name, key, secret_val)
+    yield ns, name, key
+    k8s.delete_secret(ns, name)
```
A-Hilaly, line 57: question: How/Will this secret be deleted before tests end?

A-Hilaly (approved):
LGTM 👍

I left one nit and a question below

test/e2e/tests/test_db_instance.py
jaypipes: yep, since this is a pytest fixture, it will be deleted after the smoke test that uses it completes (either successfully or fails)

PR #28, opened 2021-09-01
requeue db instance while not available

jaypipes:
After creating a new DBInstance resource, we were not setting
ConditionTypeResourceSynced on the resource nor were we asking to
requeue the resource after some time. This meant that the resource's
Status was not being updated on the Kubernetes side, even after the DB
instance's status transitioned to 'available' on the RDS side.

In this patch, we're adding two generic hook implementations for
sdk_read_many_post_set_output and sdk_create_post_set_output that will
create a ConditionTypeResourceSynced on the resource with a value of
ConditionFalse (because the resource isn't actually synced yet).  This
causes the controller to queue the resource up for a new reconciliation
loop, during which the resource's Status will be updated to the latest
observed values.

Fixes Issue aws-controllers-k8s/community#923

### DIFF
--- a/pkg/resource/db_instance/hooks.go
+++ b/pkg/resource/db_instance/hooks.go
@@ -125,6 +125,16 @@ func instanceHasTerminalStatus(r *resource) bool {
 	return false
 }
 
+// instanceAvailable returns true if the supplied DB instance is in an
+// available status
+func instanceAvailable(r *resource) bool {
+	if r.ko.Status.DBInstanceStatus == nil {
+		return false
+	}
+	dbis := *r.ko.Status.DBInstanceStatus
+	return dbis == StatusAvailable
+}
+
 // instanceCreating returns true if the supplied DB instance is in the process
 // of being created
 func instanceCreating(r *resource) bool {
--- a/templates/hooks/db_instance/sdk_create_post_set_output.go.tpl
+++ b/templates/hooks/db_instance/sdk_create_post_set_output.go.tpl
@@ -1,0 +1,9 @@
+	// We expect the DB instance to be in 'creating' status since we just
+	// issued the call to create it, but I suppose it doesn't hurt to check
+	// here.
+	if instanceCreating(&resource{ko}) {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		setSyncedCondition(&resource{ko}, corev1.ConditionFalse, nil, nil)
+		return &resource{ko}, nil
+	}
--- a/test/e2e/tests/test_db_instance.py
+++ b/test/e2e/tests/test_db_instance.py
@@ -39,6 +39,9 @@ DELETE_TIMEOUT_SECONDS = 600
 CREATE_INTERVAL_SLEEP_SECONDS = 15
 # Time to wait before we get to an expected `available` state.
 CREATE_TIMEOUT_SECONDS = 600
+# Time we wait after resource becoming available in RDS and checking the CR's
+# Status has been updated
+CHECK_STATUS_WAIT_SECONDS = 10
 
 MODIFY_WAIT_AFTER_SECONDS = 20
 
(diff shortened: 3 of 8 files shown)


test/e2e/tests/test_db_instance.py
```diff
@@ -110,6 +116,22 @@ def test_crud_postgres13_t3_micro(
...
+        # The CR's `Status.DBInstanceStatus` should be updated because the CR
+        # is requeued on successful reconciliation loops and subsequent
+        # reconciliation loops call ReadOne and should update the CR's Status
+        # with the latest observed information.
+        # https://github.com/aws-controllers-k8s/community/issues/923
+        cr = k8s.get_resource(ref)
+        assert cr is not None
+        assert 'status' in cr
+        assert 'dbInstanceStatus' in cr['status']
+        assert cr['status']['dbInstanceStatus'] != 'creating'
```
A-Hilaly, line 133: Not a big deal, but the status should normally become `available` in here
```suggestion
        assert cr['status']['dbInstanceStatus'] == 'available'
```

pkg/resource/db_instance/hooks.go
```diff
@@ -125,6 +125,16 @@ func instanceHasTerminalStatus(r *resource) bool {
...
 }
 
+// instanceAvailable returns true if the supplied DB instance is in an
+// available status
+func instanceAvailable(r *resource) bool {
+	if r.ko.Status.DBInstanceStatus == nil {
+		return false
+	}
+	dbis := *r.ko.Status.DBInstanceStatus
+	return dbis == StatusAvailable
```
vijtrip2, line 135: Would this keep requeueing if the status is not "StatusAvailable" but something else like "StatusTerminal"?
In which case we would wanna stop requeueing.

vijtrip2 (approved):
Looks good but single comment to clarify my understanding.

templates/hooks/db_instance/sdk_create_post_set_output.go.tpl
```diff
@@ -0,0 +1,9 @@
+	// We expect the DB instance to be in 'creating' status since we just
+	// issued the call to create it, but I suppose it doesn't hurt to check
+	// here.
+	if instanceCreating(&resource{ko}) {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		setSyncedCondition(&resource{ko}, corev1.ConditionFalse, nil, nil)
+		return &resource{ko}, nil
+	}
```
A-Hilaly, line 9: Looking at these, i think we should do the same for `dynamodb-controller` instead of returning a "requeueError"

pkg/resource/db_instance/hooks.go
jaypipes: Well, this particular piece of code doesn't requeue or anything.. it's just for determining if the DB instance is available or not. There are hooks for sdk_update_pre_build_request that will set the CR to a Terminal state in some cases, though:

https://github.com/aws-controllers-k8s/rds-controller/blob/main/templates/hooks/db_instance/sdk_update_pre_build_request.go.tpl

templates/hooks/db_instance/sdk_create_post_set_output.go.tpl
jaypipes: yeah, we don't want to return a requeue error in the ReadOne code path for things like that. Just adding a ConditionTypeResourceSynced condition with value of ConditionFalse is enough to trigger a requeue due to this code:

https://github.com/aws-controllers-k8s/runtime/blob/86654532edbd5e5cf99d7f89c2aa561bce810f14/pkg/runtime/reconciler.go#L610-L616

in the main reconciler.

test/e2e/tests/test_db_instance.py
jaypipes: @A-Hilaly :) originally I had it like that, but unfortunately, after `creating`, RDS sets a DBInstance to a variety of other statuses *before* `available`, including statuses like `backing-up` :)

PR #43, opened 2021-10-29
refinement of e2e tests

jaypipes:
This PR includes two patches that provide enhancements and
simplifications to the e2e testing for the RDS controller.

The first enhancement is to separate the Kubernetes secret creation
and management into a pytest fixture that is importable from the
test modules.

The second enhancement pulls the waiter and getter functionality for
DB clusters out of the test module and into a separate importable module.

This will be important for later patches where I need to test *both* DB
cluster and DB instances in the same file and don't want to copy/paste a
bunch of boilerplate waiter/getter code between test modules.

### DIFF
--- a/test/e2e/db_cluster.py
+++ b/test/e2e/db_cluster.py
@@ -1,0 +1,124 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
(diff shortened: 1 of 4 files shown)


test/e2e/db_cluster.py
```diff
@@ -0,0 +1,124 @@
...
+class StatusMatcher:
+    def __init__(self, status):
+        self.match_on = status
+
+    def __call__(self, record: dict) -> bool:
+        return 'Status' in record and record['Status'] == self.match_on
+
+
+def status_matches(status: str) -> ClusterMatchFunc:
+    return StatusMatcher(status)
```
A-Hilaly, line 42: Oooo niice!

A-Hilaly (approved):

PR #45, opened 2021-11-03
Ensure Synced Condition on DB instance/cluster

jaypipes:
Adds tests that the ACK.ResourceSynced Condition is present on resources
of kind DBCluster and DBInstance and that the status of those conditions
are either False or True depending on the expected state transition of
the DB instance or cluster.

This involved creating some helper functions in test/e2e/condition.py
that assert that a Condition of a specific type and status is present in
a resource's Status.Conditions collection.

In grep'ing through logs for the rds-controller, I noticed that the
AvailabilityZone field for the DBInstance should be late-initialized and
so added that to the generator.yaml file.

Finally, found that ACK.ResourceSynced/True was not being set on
DBCluster and DBInstance resources when the status of the instance or
cluster made its way to 'available'. I added an else block to the
generic hook code templates to ensure that ACK.ResourceSynced condition
with a status of True is present when the resource's status transitions
to 'available'.

Signed-off-by: Jay Pipes <jaypipes@gmail.com>

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -144,6 +144,8 @@ resources:
         - DBSubnetGroupNotFoundFault
         - DBParameterGroupNotFound
     fields:
+      AvailabilityZone:
+        late_initialize: {}
       DBInstanceIdentifier:
         is_primary_key: true
       MasterUserPassword:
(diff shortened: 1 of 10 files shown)


vijtrip2 (approved):
LGTM

PR #56, opened 2022-01-31
Add check for DB cluster available before updating

RedbackThomson:
Fixes: https://github.com/aws-controllers-k8s/community/issues/1107

Description of changes:
If a DB cluster is in any state other than `Available`, then an attempt to update it will return a terminal `InvalidDBClusterStateFault` error. This error will stop the cluster from reconciling any further, even though the underlying RDS DB cluster is working properly.

This PR checks that the DB cluster is in the `Available` state before running the custom update code, otherwise re-queuing it until it is.


jaypipes (approved):
Good catch, @RedbackThomson. When I added support for the customUpdate code paths, I simply forgot to fix this.

PR #61, opened 2022-02-10
Only include DBSubnetGroupName inside ModifyDBInstanceRequest when there is delta

vijtrip2:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1147

Description of changes:
* By default, ACK code-generator creates `UpdateRequest` payload completely from desired state of the resource and does not consider `delta` between `desired` & `latest` while creating the UpdateRequest payload.
* Due to above the UpdateRequest also contains the fields which are same between desired and latest state.
* This works fine in most cases where AWS Update APIs ignore the no-op updates in the backend but RDS's `ModifyDBInstance` API does not. If ModifyDBInstance API tries to update `DBSubnetGroupName` with the same name, it returns an error "InvalidVPCNetworkStateFault: The specified DB instance is already in the target DB subnet group.\n\tstatus code: 400" 
* I have added a hook that checks for the delta of `DBSubnetGroupName` before making the ModifyDBInstance API call

-----

* Tested manually and validated that the "InvalidVPCNetworkStateFault" error goes away
* added e2e test as well covering above case

### DIFF
--- a/templates/hooks/db_instance/sdk_update_post_build_request.go.tpl
+++ b/templates/hooks/db_instance/sdk_update_post_build_request.go.tpl
@@ -1,0 +1,3 @@
+	if !delta.DifferentAt("Spec.DBSubnetGroupName") {
+		input.DBSubnetGroupName = nil
+	}
--- a/templates/hooks/db_instance/sdk_update_post_build_request.go.tpl
+++ b/templates/hooks/db_instance/sdk_update_post_build_request.go.tpl
@@ -1,3 +1,8 @@
+	// ModifyDBInstance call will return ValidationError when the
+	// ModifyDBInstanceRequest contains the same DBSubnetGroupName
+	// as the DBInstance. So, if there is no delta between
+	// desired and latest for Spec.DBSubnetGroupName, exclude it
+	// from ModifyDBInstanceRequest
 	if !delta.DifferentAt("Spec.DBSubnetGroupName") {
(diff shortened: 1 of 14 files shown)


templates/hooks/db_instance/sdk_update_post_build_request.go.tpl
```diff
@@ -0,0 +1,3 @@
+	if !delta.DifferentAt("Spec.DBSubnetGroupName") {
```
jaypipes, line 1: Please put a code comment to indicate why we're doing this :) I will undoubtedly forget in the future :)

jaypipes (changes requested):
@vijtrip2 nice sleuthing deep dive on this one! Reminds me of why [I had to create a custom update method for DBCluster](https://github.com/aws-controllers-k8s/rds-controller/blob/0465deb6e26c5593574c3ac8fa62622a793255c8/pkg/resource/db_cluster/custom_update.go#L30-L33) (when `engine_mode` is `serverless`, the behaviour of the ModifyDBCluster operation is different from when the `engine_mode` is `provisioned` -- the API rejects values in the input shape that are the same as the current value. But only when `serverless`.... *sigh*).

One request for a code comment in the template, though. Otherwise, this is fantastic. Thank you!

jaypipes (approved):
Thanks for adding that code comment @vijtrip2 :)

PR #65, opened 2022-03-05
handle asynchronous updates for DBInstance by requeueing

vijtrip2:
Issue [#1156](https://github.com/aws-controllers-k8s/community/issues/1156)

Description of changes:
* During DBInstance update, when `ModifyDBInstance` call returns success then the ACK reconciler marks the resource as Synced and the reconciliation stops. However asynchronous update of DBInstance happens on server side and the ACK reconciler does not sync them.
* Add two new hooks, one to requeue after successful update and another to handle delta for non-late-initialized fields. 
* I did not use the `requeueOnSuccessSeconds` construct because it will cause the reconciliation to go forever and for some amount of time, the resource condition will be synced, even when the DBInstance is in modifying state.
* Do not set Synced condition true inside post_read_many hook, but depend on `ensureConditions` method

**Test Updates**
* Update test manifest to use smaller storage to improve the test time and use engine version 14.1 because engine version 13.2 was throwing unsupported errors
* Added e2e test for 'MultiAZ' field update as mentioned in the issue [#1156](https://github.com/aws-controllers-k8s/community/issues/1156)
* Updated the test to wait for Synced condition and then assert DBInstance fields using AWS API calls


pkg/resource/db_instance/hooks.go
```diff
@@ -90,6 +90,13 @@ var (
 		errors.New("DB instance in 'deleting' state, cannot be modified or deleted."),
 		ackrequeue.DefaultRequeueAfterDuration,
 	)
+
+	customWaitAtferUpdate = ackrequeue.NeededAfter(
+		errors.New("ModifyDBInstance API was successful but it"+
+			" asynchronously updates the DBInstanceStatus. Requeue to find the"+
+			" current DBInstance status"),
```
jaypipes, line 97: Please change the message to inform the user that they should not take any action. The imperative "Requeue to find the current DBInstance status" makes it sound like the user should take an action...

Suggest rewording to this:

`Requeueing resource after successful update; status fields retrieved asynchronously."

templates/hooks/db_instance/delta_pre_compare.go.tpl
```diff
@@ -0,0 +1,12 @@
...
+    // desired(a) and is present in latest(b) because AvailabilityZone field is
+    // late-initialized
+    // This special handling is only needed for DBInstance because a requeue
+    // error needs to be returned even after successful ModifyDBInstance call.
+    // See sdk_update_post_set_output.go.tpl for more details.
+    // The requeue error from update prevents the late initialization in
+    // reconciler.go and without ignoring AvailabilityZone for delta , the
+    // reconciler will keep updating the dbinstance and constantly requeueing it.
+    if a != nil && a.ko.Spec.AvailabilityZone == nil && b != nil && b.ko.Spec.AvailabilityZone != nil {
+        a.ko.Spec.AvailabilityZone = b.ko.Spec.AvailabilityZone
```
jaypipes, line 11: isn't this what lateInitialize is for?

jaypipes (changes requested):
Couple questions inline, Vijay...

templates/hooks/db_instance/delta_pre_compare.go.tpl
vijtrip2: Yep, but when the requeue error is returned from the update method lateInitialization is skipped.

If the above hook is not present then following steps occur:
Update -> Requeue -> Delta found for AZ field -> Update -> Requeue -> Delta found for AZ

With the above hook the following steps occur:
Update -> Requeue -> No Delta found for AZ field -> No Update -> LateInitialize -> Synced False -> Requeue -> No Delta -> LateInitialize ...

pkg/resource/db_instance/hooks.go
```diff
@@ -90,6 +90,12 @@ var (
 		errors.New("DB instance in 'deleting' state, cannot be modified or deleted."),
 		ackrequeue.DefaultRequeueAfterDuration,
 	)
+
+	customWaitAtferUpdate = ackrequeue.NeededAfter(
+		errors.New("Requeueing resource after successful update; status fields"+
```
jaypipes, line 95: needs a space after "fields" and before "retrieved" :)

templates/hooks/db_instance/delta_pre_compare.go.tpl
jaypipes: I see... seems like a better long-term solution would be to have `lateInitialize` execute when a requeue error is returned from `Update`, no?

vijtrip2:
> seems like a better long-term solution would be to have lateInitialize execute when a requeue error is returned from Update, no?

Yes. Definitely. Requeue errors can be handled differently than other errors for create/update. I did not wish to include runtime change as part of this PR so did not go down that route.

templates/hooks/db_instance/delta_pre_compare.go.tpl
vijtrip2: Yes. Definitely. Requeue errors can be handled differently than other errors for create/update. I did not wish to include runtime change as part of this PR so did not go down that route.

PR #78, opened 2022-05-26
RDS Create DBInstance From Snapshot 

brucegucode:
Fixes https://github.com/aws-controllers-k8s/community/issues/1262 

Description of changes:

This change is allow customer create db instance from DBSnapshotIdentifier when it's present. 

When DBSnapshotIdentifier is present in yaml file, ACK will try call [RestoreDBInstanceFromDBSnapshot](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBInstanceFromDBSnapshot.html#API_RestoreDBInstanceFromDBSnapshot) API instead of normal CreateDbInstance api. 

Tested this new behavior by adding `dbSnapshotIdentifier: "arn:aws:rds:us-west-2:980401675890:snapshot:brucegu-snapshot-1"` in yaml file and make sure DBInstance can be created from this snapshot. 


pkg/resource/db_instance/hooks.go
```diff
@@ -14,10 +14,27 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Hack to avoid import errors during build...
+var (
+	_ = &metav1.Time{}
+	_ = &svcsdk.RDS{}
+	_ = &svcapitypes.DBInstance{}
+	_ = ackv1alpha1.AWSAccountID("")
+	_ = &ackcondition.NotManagedMessage
```
RedbackThomson, line 37: I don't think this is necessary for hook files. We can optionally import if we need these packages.

pkg/resource/db_instance/hooks.go
```diff
@@ -160,3 +177,826 @@ func instanceDeleting(r *resource) bool {
...
+	rm.metrics.RecordAPICall("CREATE", "RestoreDbInstanceFromDbSnapshot", respErr)
+	if respErr != nil {
+		return nil, respErr
+	}
+
+	// Merge in the information we read from the API call above to the copy of
+	// the original Kubernetes object we passed to the function
+	ko := r.ko.DeepCopy()
+
+	if resp.DBInstance.ActivityStreamEngineNativeAuditFieldsIncluded != nil {
```
RedbackThomson, line 206: I will message you about how we can make the code-generator generate all of this ugly code

pkg/resource/db_instance/hooks.go
```diff
@@ -160,3 +177,826 @@ func instanceDeleting(r *resource) bool {
...
+}
+
+// restoreDbInstanceFromDbSnapshotPayload returns an SDK-specific struct for the HTTP request
+// payload of the RestoreDbInstanceFromDbSnapshot API call
+func (rm *resourceManager) restoreDbInstanceFromDbSnapshotPayload(
+	r *resource,
+) (*svcsdk.RestoreDBInstanceFromDBSnapshotInput, error) {
+	res := &svcsdk.RestoreDBInstanceFromDBSnapshotInput{}
+
+	if r.ko.Spec.DBSnapshotIdentifier != nil {
```
RedbackThomson, line 879: Likewise we should be able to auto-generate this

RedbackThomson (changes requested):
Nice. This is exactly how I envisioned it when we first talked about it. I'll reach out and we can walk through how to reduce the repetitive code since we really don't want to maintain that.

As for tests, is there a reliable way for us to run integration tests for this? I suppose there isn't a way to create a DBSnapshot without first creating a normal RDS DB and then calling `CreateDBSnapshot`? I'm trying to think about how we would do that for CI

brucegucode:
synced with @RedbackThomson offline and added new template sdk_file_end.go.tpl to remove unnecessary get/set field code in commit 7 inside this PR

pkg/resource/db_instance/hooks.go
brucegucode: synced offline, addressed in next revision

pkg/resource/db_instance/hooks.go
brucegucode: synced offline, addressed in next revision

pkg/resource/db_instance/hooks.go
brucegucode: Thanks Nick!

pkg/resource/db_instance/hooks.go
```diff
@@ -160,3 +164,33 @@ func instanceDeleting(r *resource) bool {
...
+	rm.setResourceFromRestoreDBInstanceFromDBSnapshotOutput(r, resp)
+	rm.setStatusDefaults(r.ko)
+
+	// We expect the DB instance to be in 'creating' status since we just
+	// issued the call to create it, but I suppose it doesn't hurt to check
+	// here.
+	if instanceCreating(&resource{r.ko}) {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		ackcondition.SetSynced(&resource{r.ko}, corev1.ConditionFalse, nil, nil)
```
vijtrip2, line 192: nit: for setting synced condition, I will recommend "synced" construct inside generator config(generator.yaml)

brucegucode: got it, Thanks!

PR #82, opened 2022-06-07
Issue 1251

brucegucode:
Issue #, if available:
both
https://github.com/aws-controllers-k8s/community/issues/1265
and
https://github.com/aws-controllers-k8s/community/issues/1251

Description of changes:
1. For issue 1251, the root cause is found an addressed here https://github.com/aws-controllers-k8s/runtime/pull/92, so this pr will remove the unnecessary rename PI part, also we don't need to rename for other fields like enableIAM, enableCloudWatch as well. 

2. During engine version testing, 1265 is a low hanging fruit, we just need override AllowMajorVersionUpgrade to true so customer can do major version upgrade. This flag AllowMajorVersionUpgrade was designed to protect instance from accidentally upgrade , but it's not applying to controller, we can override it to true as always. 

### DIFF
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -179,6 +179,10 @@ resources:
         from:
           operation: RestoreDBInstanceFromDBSnapshot
           path: UseDefaultProcessorFeatures
+      PerformanceInsightsKMSKeyID:
+        late_initialize: {}
+      PerformanceInsightsRetentionPeriod:
+        late_initialize: {}
     renames:
       operations:
         CreateDBInstance:
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -35,6 +35,11 @@ operations:
       # makes to a resource's Spec to be reconciled by the ACK service
       # controller, not a different service.
       ApplyImmediately: true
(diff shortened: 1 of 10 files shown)


apis/v1alpha1/generator.yaml
```diff
@@ -179,14 +189,6 @@ resources:
...
           operation: RestoreDBInstanceFromDBSnapshot
           path: UseDefaultProcessorFeatures
-    renames:
-      operations:
-        CreateDBInstance:
-          input_fields:
-            EnablePerformanceInsights: PerformanceInsightsEnabled
-        ModifyDBInstance:
-          input_fields:
-            EnablePerformanceInsights: PerformanceInsightsEnabled
```
RedbackThomson, line 189: This change is still necessary. Even if it doesn't break functionality, we want to reduce duplication between fields in the spec and status. Since this property of `DBInstance` is configurable (it isn't read-only), it should appear only in the spec. We should do the same for the other properties we identified as well.

RedbackThomson (changes requested):
Left an inline comment about the necessity of the changes.

Could you also please rename the PR to match the functional changes of the code, and instead only refer to the issue you're resolving in the body of the PR description. Thanks

PR #83, opened 2022-06-08
Allow major version upgrade by setting AllowMajorVersionUpgrade to true

brucegucode:
Fixes aws-controllers-k8s/community#1265

Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1265

Description of changes:
This PR is setting AllowMajorVersionUpgrade to true for both modify db instance and modify db cluster.
This flag was designed as protect flag to avoid accidentally engine version upgrade, however this is not necessary for controller side. whenever customer make change to engine version, ACK controller should take instance to target engine version. 

### DIFF
--- a/pkg/resource/db_cluster/custom_update.go
+++ b/pkg/resource/db_cluster/custom_update.go
@@ -542,6 +542,7 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 	res := &svcsdk.ModifyDBClusterInput{}
 
 	res.SetApplyImmediately(true)
+	res.SetAllowMajorVersionUpgrade(true)
 	if r.ko.Spec.BacktrackWindow != nil && delta.DifferentAt("Spec.BacktrackWindow") {
 		res.SetBacktrackWindow(*r.ko.Spec.BacktrackWindow)
 	}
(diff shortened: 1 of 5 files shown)


pkg/resource/db_cluster/custom_update.go
```diff
@@ -542,6 +542,7 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 	res := &svcsdk.ModifyDBClusterInput{}
 
 	res.SetApplyImmediately(true)
+	res.SetAllowMajorVersionUpgrade(true)
```
A-Hilaly, line 545: Updating `AllowMajorVersionUpgrade` is only allowed for Amazon Aurora DB Clusters. I supposed the API will reject all the update operations for non Aurora DBclusters

brucegucode: I think the message  `Valid for: Aurora DB clusters only` was kind of outdated. aurora cluster is the first cluster concept introduced in rds, later when global db cluster and multi-az cluster comes in, the doc is not updated.
if you see modify global cluster, this flag is also available. https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyGlobalCluster.html

A-Hilaly (approved):
:+1:

jaypipes (approved):
:+1: thanks @brucegucode!

PR #84, opened 2022-06-09
RDS: Add support for create DBInstanceReadReplica

brucegucode:
**Issue #**, if available: issue-893 

**Description of changes:**
https://github.com/aws-controllers-k8s/community/issues/893

This pr is adding ability to allow create DBInstanceReadReplica, api call: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstanceReadReplica.html

**Testing part:** 
Tested with successful creation of 
- same region + encrypted source instance
- same region + unencrypted source instance
- cross region + encrypted source instance
- cross region + unencrypted source instance

**Notes:** 
I don't have e2e test for this case yet which is because I'm blocked on this issue. https://github.com/aws-controllers-k8s/community/issues/1325 
Will add e2e tests after 1325 issue is resolved. 

Also this pr has a dependency pr: https://github.com/aws-controllers-k8s/code-generator/pull/339 


jaypipes:
Holding until the v0.19.0 PR lands

apis/v1alpha1/generator.yaml
```diff
@@ -183,6 +184,31 @@ resources:
...
+          path: DestinationRegion
+      ReplicaMode:
+        from:
+          operation: CreateDBInstanceReadReplica
+          path: ReplicaMode
+      SourceRegion:
+        from:
+          operation: CreateDBInstanceReadReplica
+          path: SourceRegion
+      PreSignedUrl:
```
RedbackThomson, line 204: Please capitalise to `URL`

apis/v1alpha1/generator.yaml
```diff
@@ -183,6 +184,31 @@ resources:
...
+        from:
+          operation: CreateDBInstanceReadReplica
+          path: PreSignedUrl
+      EnablePerformanceInsights:
```
RedbackThomson, line 208: `PerformanceInsightsEnabled` already exists on `DBInstance`? Why do we need this as well?

pkg/resource/db_instance/hooks.go
```diff
@@ -194,3 +194,33 @@ func (rm *resourceManager) restoreDbInstanceFromDbSnapshot(
...
+	// We expect the DB instance to be in 'creating' status since we just
+	// issued the call to create it, but I suppose it doesn't hurt to check
+	// here.
+	if instanceCreating(&resource{r.ko}) {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		ackcondition.SetSynced(&resource{r.ko}, corev1.ConditionFalse, nil, nil)
+		return &resource{r.ko}, nil
+	}
+	return &resource{r.ko}, nil
```
RedbackThomson, line 225: Can probably remove the return within the condition since these are duplicated

templates/hooks/db_instance/sdk_create_pre_build_request.go.tpl
```diff
@@ -4,3 +4,9 @@
     	created, err = rm.restoreDbInstanceFromDbSnapshot(ctx, desired)
         return created, err
     }
+    // if request has SourceDBInstanceIdentifier spec, create request will call CreateDBInstanceReadReplicaWithContext
+    // instead of normal create api
+    if desired.ko.Spec.SourceDBInstanceIdentifier != nil {
+        created, err = rm.createDBInstanceReadReplica(ctx, desired)
+        return created, err
```
RedbackThomson, line 11: nit:
```suggestion
        return rm.createDBInstanceReadReplica(ctx, desired)
```

templates/hooks/db_instance/sdk_file_end.go.tpl
```diff
@@ -1,7 +1,9 @@
 {{ $CRD := .CRD }}
 {{ $SDKAPI := .SDKAPI }}
 
-{{- $operation := (index $SDKAPI.API.Operations "RestoreDBInstanceFromDBSnapshot")}}
+{{ range $operationName := Each "RestoreDBInstanceFromDBSnapshot" "CreateDBInstanceReadReplica" }}
```
RedbackThomson, line 4: ++ clean

RedbackThomson (changes requested):
Happy for this to go ahead, just please take a look at your spec fields so we don't end up with duplicates

pkg/resource/db_instance/hooks.go
brucegucode: you're right, will remove it

apis/v1alpha1/generator.yaml
brucegucode: looks like RDS api take it as `Url` instead of `URL` e.g. https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance-read-replica.html

apis/v1alpha1/generator.yaml
brucegucode: I think it's because in DBInstance, `EnablePerformanceInsights` is already renamed to `PerformanceInsightsEnabled`,  controller will throw exception if i don't add it here. 


```
# github.com/aws-controllers-k8s/rds-controller/pkg/resource/db_instance
pkg/resource/db_instance/sdk.go:3684:15: r.ko.Spec.EnablePerformanceInsights undefined (type "github.com/aws-controllers-k8s/rds-controller/apis/v1alpha1".DBInstanceSpec has no field or method EnablePerformanceInsights)
pkg/resource/db_instance/sdk.go:3685:47: r.ko.Spec.EnablePerformanceInsights undefined (type "github.com/aws-controllers-k8s/rds-controller/apis/v1alpha1".DBInstanceSpec has no field or method EnablePerformanceInsights)

```

templates/hooks/db_instance/sdk_create_pre_build_request.go.tpl
brucegucode: makes sense

templates/hooks/db_instance/sdk_file_end.go.tpl
brucegucode: can you elaborate a little bit?

brucegucode:
i'm deprecating this pr as now i'm using my own forked repo and it's tracked here https://github.com/aws-controllers-k8s/rds-controller/pull/92

apis/v1alpha1/generator.yaml
RedbackThomson: I understand that, but we transform all of our Spec and Status field initialisms into full upper case. For example `ARN` is always upper-case. Our code generator knows to transform the casing when creating AWS SDK types

apis/v1alpha1/generator.yaml
RedbackThomson: In the same way you rename the field for the Create call, you should do it with this field.

templates/hooks/db_instance/sdk_file_end.go.tpl
RedbackThomson: Just saying this code is clean. No action here

apis/v1alpha1/generator.yaml
brucegucode: I think i'm following the same pattern for `CreateDBInstance` call as this `CreateDBInstanceReadReplica` call 

I have this rename here https://github.com/aws-controllers-k8s/rds-controller/blob/issue-893/generator.yaml#L220-L222 .

it's strange that if i remove the line 208-211 https://github.com/aws-controllers-k8s/rds-controller/blob/issue-893/generator.yaml#L208-L211.  

generator will throw above exception.

apis/v1alpha1/generator.yaml
brucegucode: i see, will use capital case here

PR #91, opened 2022-06-20
Ignore NetworkType param if no change, not consider diff when engine version diff in minor version

brucegucode:
Issue #, if available:

no issue number 

Description of changes:
2 changes in this pr

1. ignore NetworkType param when no delta found. RDS coral service will accept request when param has NetworkType param and trigger workflow, although inside rds workflow, it will do noop. it's still not best experience for customer since instance will leave active/none state when this happen.  (this is introduced recently by updating rds api version)

3. RDS will choose preferred engine minor version when only major version give. controller should not try to take them as different and try to sync them. e.g. spec has `14`, rds status shows `14.1`, they should be treated same. Before this fix, controller is trying to sync them forever.  Previous error log: 
```
1.6557458184530575e+09	INFO	ackrt	desired resource state has changed	{"account": "980401675890", "role": "", "region": "us-west-2", "kind": "DBInstance", "namespace": "default", "name": "brucegu-test-tag-2", "is_adopted": false, "generation": 3, "diff": [{"Path":{"Parts":["Spec","EngineVersion"]},"A":"14","B":"14.1"}]}
```

### DIFF
--- a/pkg/resource/db_instance/hooks.go
+++ b/pkg/resource/db_instance/hooks.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
 	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
@@ -193,4 +194,17 @@ func (rm *resourceManager) restoreDbInstanceFromDbSnapshot(
 		return &resource{r.ko}, nil
 	}
 	return &resource{r.ko}, nil
+}
+
+// RDS will choose preferred engine minor version if only
+// engine major version is provided and controler should not
+// treat them as different, sunch as spec has 14, status has 14.1
+// controller should treat them as same
+func reconcileEngineVersion(
+	a *resource,
+	b *resource,
+) {
+	if a != nil && b != nil && strings.HasPrefix(*b.ko.Spec.EngineVersion, *a.ko.Spec.EngineVersion) {
+		a.ko.Spec.EngineVersion = b.ko.Spec.EngineVersion
+	}
 }
--- a/pkg/resource/db_instance/hooks.go
+++ b/pkg/resource/db_instance/hooks.go
@@ -198,7 +198,7 @@ func (rm *resourceManager) restoreDbInstanceFromDbSnapshot(
 
 // RDS will choose preferred engine minor version if only
 // engine major version is provided and controler should not
-// treat them as different, sunch as spec has 14, status has 14.1
+// treat them as different, such as spec has 14, status has 14.1
 // controller should treat them as same
 func reconcileEngineVersion(
 	a *resource,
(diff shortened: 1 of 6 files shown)


pkg/resource/db_instance/hooks.go
```diff
@@ -194,3 +195,16 @@ func (rm *resourceManager) restoreDbInstanceFromDbSnapshot(
...
+// engine major version is provided and controler should not
+// treat them as different, sunch as spec has 14, status has 14.1
+// controller should treat them as same
+func reconcileEngineVersion(
+	a *resource,
+	b *resource,
+) {
+	if a != nil && b != nil && strings.HasPrefix(*b.ko.Spec.EngineVersion, *a.ko.Spec.EngineVersion) {
+		a.ko.Spec.EngineVersion = b.ko.Spec.EngineVersion
+	}
```
RedbackThomson, line 209: I think this logic will fail if you set the minor version in the spec. For instance, if you set `Spec = 14.1` and `Status = 14.0`, this will not upgrade the cluster version. Is that expected?

pkg/resource/db_instance/hooks.go
```diff
@@ -194,3 +195,16 @@ func (rm *resourceManager) restoreDbInstanceFromDbSnapshot(
 	}
 	return &resource{r.ko}, nil
 }
+
+// RDS will choose preferred engine minor version if only
+// engine major version is provided and controler should not
+// treat them as different, sunch as spec has 14, status has 14.1
```
RedbackThomson, line 201:
```suggestion
// treat them as different, such as spec has 14, status has 14.1
```

RedbackThomson (commented):
If you wouldn't mind just double checking the logic in the comparison, then I'll be happy to push this through. Thanks

pkg/resource/db_instance/hooks.go
brucegucode: I think it's working. like `strings.HasPrefix("14.1", "14,0")` is returning false , so controller will not do the override here and treat them as different

spec file has 14.2 and then i modified it to 14.3 and it's getting the diff.
`1.6557526651883025e+09	INFO	ackrt	desired resource state has changed	{"account": "980401675890", "role": "", "region": "us-west-2", "kind": "DBInstance", "namespace": "default", "name": "brucegu-test-tag-3", "is_adopted": false, "generation": 11, "diff": [{"Path":{"Parts":["Spec","EngineVersion"]},"A":"14.3","B":"14.2"}]}`

pkg/resource/db_instance/hooks.go
brucegucode: updated in next revision

pkg/resource/db_instance/hooks.go
RedbackThomson: Oh oh I see. You're using `strings.HasPrefix`. Yep that makes sense

vijtrip2:
Would you consider updating the e2e manifest to validate this change?

/approve

brucegucode:
> how about updating the e2e manifest to validate this change?
> 
> /approve

I'm still in process of upgrading the bootstrap logic in rds controller.  will add bunch of test validate/new tests after that including validating this commit.

PR #93, opened 2022-06-22
v0.0.26 release

jkatz:
Release artificates for v0.0.26

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,8 +1,8 @@
 ack_generate_info:
-  build_date: "2022-06-21T21:15:15Z"
-  build_hash: a45f3b900849ec03c5e16ed2778c0b8e2923ffee
-  go_version: go1.18.2
-  version: v0.19.1
+  build_date: "2022-06-22T20:33:56Z"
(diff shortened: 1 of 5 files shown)


apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,8 +1,8 @@
 ack_generate_info:
-  build_date: "2022-06-21T21:15:15Z"
-  build_hash: a45f3b900849ec03c5e16ed2778c0b8e2923ffee
-  go_version: go1.18.2
-  version: v0.19.1
+  build_date: "2022-06-22T20:28:52Z"
+  build_hash: 7669f36bb0005c47869464a22f30193ea42793ca
+  go_version: go1.18.3
+  version: v0.19.0-6-g7669f36
```
RedbackThomson, line 5: I see a regression in your `code-generator` version. Please update that

jkatz: That was the issue -- didn't fetch the tags. Repushed.

RedbackThomson: Oh yes that'll do it. Really we should be doing a `git fetch upstream` as part of `make build-controller` to negate this

PR #95, opened 2022-06-24
Fix `DBInstance` for updating PI and add tests

RedbackThomson:
Fixes https://github.com/aws-controllers-k8s/community/issues/1346

Description of changes:
Updating `EnablePerformanceInsights` caused the `DBInstance` to cycle continuously, as it tried to reconcile the server side default values and handle the asynchronous status updates. Rather than relying on a requeue error to wait for the status to update, this pull request modifies the `DBInstance` custom code to use `ResourceSynced = false`. Furthermore, we remove the server side defaulted values from the diff if they have not been updated in the desired spec.


vijtrip2:
good work!
/approve

generator.yaml
```diff
@@ -168,13 +168,20 @@ resources:
         is_immutable: true
       DBInstanceIdentifier:
         is_primary_key: true
+      DBInstanceStatus:
+        print:
+          name: "STATUS"
```
brucegucode, line 173: just curious what is the purpose of this DBInstanceStatus here

go.local.mod
```diff
@@ -0,0 +1,71 @@
+module github.com/aws-controllers-k8s/rds-controller
```
brucegucode, line 1: also curious what is the function of this file ?

brucegucode (commented):
Looks good to me! just a few curiosity question

go.local.mod
RedbackThomson: Line 5 of this file indicates that we should use the locally cloned version of runtime instead of the dependency from Github. This makes it easier to debug runtime changes, because I don't need to have the changes committed. I've had this file locally for a while, when debugging `v0.19.0`, so I figured I'd just push these up here in case anyone else needs them. For the most part you won't need to touch these.

brucegucode: got it

PR #96, opened 2022-07-05
Add DBCluster serverlessV2ScalingConfiguration update support

brucegucode:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1373

Description of changes:

2 changes here + auto build which takes latest runtime changes in rds

1.  rename the scalingConfiguration to scalingConfigurationInfo in Create/ModifyDBClusterInput, so it only appear in Spec instead of Status of cluster.
2. fix issue of 1373, it is due to we currently still use custom update logic, and need update it based on rds api update.  see [pkg/resource/db_cluster/custom_update.go](https://github.com/aws-controllers-k8s/rds-controller/compare/main...brucegucode:rds-controller:issue-1373?expand=1#diff-1f439d0df6b144c73ac51eff110e99fe2262d68871d881f6c6eda8db5bfbe9b1) for detail 

Testing:
Verified by updating serverlessV2ScalingConfiguration minCap or maxCap and verified it got applied on rds end and k8s object is in sync with rds end. 

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -110,11 +110,11 @@ resources:
     renames:
       operations:
         CreateDBCluster:
-          input_fields:
-            ScalingConfiguration: ScalingConfigurationInfo
+          output_fields:
+            ScalingConfigurationInfo: ScalingConfiguration
         ModifyDBCluster:
-          input_fields:
-            ScalingConfiguration: ScalingConfigurationInfo
+          output_fields:
+            ScalingConfigurationInfo: ScalingConfiguration
   DBClusterParameterGroup:
     renames:
       operations:
--- a/generator.yaml
+++ b/generator.yaml
@@ -107,6 +107,14 @@ resources:
           resource: Key
           service_name: kms
           path: Status.ACKResourceMetadata.ARN
+    renames:
+      operations:
+        CreateDBCluster:
+          input_fields:
+            ScalingConfiguration: ScalingConfigurationInfo
+        ModifyDBCluster:
+          input_fields:
+            ScalingConfiguration: ScalingConfigurationInfo
   DBClusterParameterGroup:
     renames:
       operations:
--- a/pkg/resource/db_cluster/custom_update.go
+++ b/pkg/resource/db_cluster/custom_update.go
@@ -611,13 +611,13 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 	// For ServerlessV2ScalingConfiguration, MaxCapacity and MinCapacity,  both need appear in modify call to get ServerlessV2ScalingConfiguration modified
 	if r.ko.Spec.ServerlessV2ScalingConfiguration != nil && delta.DifferentAt("Spec.ServerlessV2ScalingConfiguration") {
 		f23 := &svcsdk.ServerlessV2ScalingConfiguration{}
-		if r.ko.Spec.ServerlessV2ScalingConfiguration.MaxCapacity != nil && r.ko.Spec.ServerlessV2ScalingConfiguration.MinCapacity != nil && delta.DifferentAt("Spec.ServerlessV2ScalingConfiguration.MaxCapacity") {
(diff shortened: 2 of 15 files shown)


pkg/resource/db_cluster/custom_update.go
```diff
@@ -630,6 +608,16 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 		}
 		res.SetVpcSecurityGroupIds(f23)
 	}
-
+        // For ServerlessV2ScalingConfiguration, MaxCapacity and MinCapacity,  both need appear in modify call to be modified
```
RedbackThomson, line 611: I think it would actually be better if we separated them. Here is my logic:

If a user only has `MinCapacity` (and not `MaxCapacity`), then this code will silently ignore the scalingconfiguration change and not make the modify call. But everything else will appear to succeed. The controller will then be stuck in a loop as it describes a different `MinCapacity` to the one in the spec.

If we add them independently of the other, then if a user specifies only `MinCapacity`, the `ModifyDBCluster` call will fail. The error it returns will be populated into the condition of the object and it (hopefully) will be obvious to the user that they need to add `MaxCapacity`. Also, returning an error will pause reconciliation (with a `Terminal` condition) until it is updated.

RedbackThomson (changes requested):

pkg/resource/db_cluster/custom_update.go
brucegucode: That's a good point! it makes sense to me

generator.yaml
```diff
@@ -107,6 +107,14 @@ resources:
           resource: Key
           service_name: kms
           path: Status.ACKResourceMetadata.ARN
+    renames:
+      operations:
+        CreateDBCluster:
+          input_fields:
+            ScalingConfiguration: ScalingConfigurationInfo
```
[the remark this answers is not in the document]
brucegucode: sure, i will adjust the output field, the naming ScalingConfiguration sounds more correct

pkg/resource/db_cluster/custom_update.go
```diff
@@ -608,16 +608,18 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
...
-                                f23.SetMinCapacity(*r.ko.Spec.ServerlessV2ScalingConfiguration.MinCapacity)
-                        }
-                }
-                res.SetServerlessV2ScalingConfiguration(f23)
+		if r.ko.Spec.ServerlessV2ScalingConfiguration.MaxCapacity != nil && delta.DifferentAt("Spec.ServerlessV2ScalingConfiguration.MaxCapacity") {
+			f23.SetMaxCapacity(*r.ko.Spec.ServerlessV2ScalingConfiguration.MaxCapacity)
+			f23.SetMinCapacity(*r.ko.Spec.ServerlessV2ScalingConfiguration.MinCapacity)
+		}
+		if r.ko.Spec.ServerlessV2ScalingConfiguration.MinCapacity != nil && delta.DifferentAt("Spec.ServerlessV2ScalingConfiguration.MinCapacity") {
+			f23.SetMaxCapacity(*r.ko.Spec.ServerlessV2ScalingConfiguration.MaxCapacity)
```
RedbackThomson, line 619: This will cause a panic if `r.ko.Spec.ServerlessV2ScalingConfiguration == nil`. Please remove lines 616 and 619

brucegucode: Because the new api `ServerlessV2ScalingConfiguration` need both MinCapacity and MaxCapacity to appear, so whenever min or max get changed, we have to add both param to `f23`. 

I'm adding a nil check for both min and max in next revision and in either case, we set both min/max value at same time. 
With some more thinking, I think it may not be able to avoid the problem that silently fail when one val is set as nil.

vijtrip2 (approved):

brucegucode:
the failed test is due to https://github.com/aws-controllers-k8s/community/issues/1376, which is fliping flop of synced state.

PR #99, opened 2022-07-11
When db instance is inside db cluster(when DBClusterIdentifier appear), set cluster level params to nil 

brucegucode:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1028

Description of changes:
When db instance is inside a db cluster, it's either an aurora case or multi-az cluster case. There are some params that should not be controlled in instance level and should go to cluster level.
So when `DBClusterIdentifier` appear, set those values to nil so runtime won'r try to modify them and get RDS `InvalidParameterCombination`. 

Testing:
1. Create a DBCluster first 
2. Create a dbinstance in it using DBClusterIdentifier param, like below, verify create succeed 
```
kind: DBInstance
metadata:
  name: "brucegu-manfred-instance-0"
spec:
  dbInstanceClass: db.r5.large
  engine: aurora-postgresql
  dbInstanceIdentifier: "brucegu-manfred-instance-0"
  performanceInsightsEnabled: true
  copyTagsToSnapshot: true
  dbClusterIdentifier: "brucegu-test-cluster-3"
```
3. Try change one of the param, like `copyTagsToSnapshot` and verify rds will throw exception like `The specified DB Instance is a member of a cluster. Modify backup window for the DB Cluster using the ModifyDbCluster API` 
4. With new code, make sure modify can go through and no exception is thrown. 

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,7 +1,7 @@
 ack_generate_info:
-  build_date: "2022-07-11T19:31:21Z"
+  build_date: "2022-07-11T21:43:34Z"
   build_hash: d09bac5fa87cffd0028014833a5e7e786c0187dd
-  go_version: go1.18.3
+  go_version: go1.18.2
   version: v0.19.2-6-gd09bac5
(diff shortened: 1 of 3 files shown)


brucegucode:
test failure is still the issue of https://github.com/aws-controllers-k8s/community/issues/1376

PR #100, opened 2022-07-12
Fix issue of adopted resource for both DBClusterParameterGroup and DBSubnetGroup

brucegucode:
Issue #, if available:

https://github.com/aws-controllers-k8s/community/issues/1389

Description of changes:
Fix issue of adopted resource for both DBClusterParameterGroup and DBSubnetGroup
1. for DBClusterParameterGroup, it's due to the rename issue that `DBParameterGroupFamily` need be called `Family` and `DBClusterParameterGroupName` need be called `Name`
2. for DBSubnetGroup, synced with @RedbackThomson , it's due to rds output status of `DBSubnetGroup` doesn't have `SubnetIDs`, and it only has `Subnets. SubnetIdentifier`, so we need to use hook to append each of them. 

With these above fixed, adopted resource should work for all available RDS CRDs now. 

### DIFF
--- a/pkg/resource/db_instance/sdk.go
+++ b/pkg/resource/db_instance/sdk.go
@@ -1775,7 +1775,6 @@ func (rm *resourceManager) sdkUpdate(
 	// For dbInstance inside dbCluster, it's either aurora or
 	// multi-az cluster case, in either case, the below params
 	// are not controlled in instance level.
-	// multi-az cluster control it on cluster leve, aurora doesn't need it
 	// hence when DBClusterIdentifier appear, set them to nil
 	// Please refer to doc : https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DeleteDBInstance.html
 	if desired.ko.Spec.DBClusterIdentifier != nil {
--- a/templates/hooks/db_subnet_group/sdk_read_many_post_set_output.go.tpl
+++ b/templates/hooks/db_subnet_group/sdk_read_many_post_set_output.go.tpl
@@ -6,3 +6,11 @@
         }
         ko.Spec.Tags = tags
 	}
+
+        if ko.Status.Subnets != nil {
+            for _, subnetIdIter := range ko.Status.Subnets {
+                if subnetIdIter.SubnetIdentifier != nil {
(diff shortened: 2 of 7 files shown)


templates/hooks/db_subnet_group/sdk_read_many_post_set_output.go.tpl
```diff
@@ -6,3 +6,11 @@
...
         ko.Spec.Tags = tags
 	}
+
+        if ko.Status.Subnets != nil {
+            for _, subnetIdIter := range ko.Status.Subnets {
+                if subnetIdIter.SubnetIdentifier != nil {
+                    ko.Spec.SubnetIDs = append(ko.Spec.SubnetIDs, subnetIdIter.SubnetIdentifier)
+                }
+            }
+        }
```
RedbackThomson, line 16: Indentation looks a little off here?

pkg/resource/db_instance/sdk.go
```diff
@@ -1775,7 +1775,6 @@ func (rm *resourceManager) sdkUpdate(
 	// For dbInstance inside dbCluster, it's either aurora or
 	// multi-az cluster case, in either case, the below params
 	// are not controlled in instance level.
-	// multi-az cluster control it on cluster leve, aurora doesn't need it
```
RedbackThomson, line 1778: Any idea what this is?

templates/hooks/db_subnet_group/sdk_read_many_post_set_output.go.tpl
brucegucode: yeah, fixed in next revision

pkg/resource/db_instance/sdk.go
brucegucode: looks like it's a build/merge issue, this comment is removed in template  https://github.com/aws-controllers-k8s/rds-controller/blob/main/templates/hooks/db_instance/sdk_update_post_build_request.go.tpl#L30 and should be removed here as well

jaypipes (approved):
:+1: good work on this @brucegucode. Test failures are unrelated... merging this.

PR #105, opened 2022-08-01
Support RDSProxy CRD

brucegucode:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/1298

Description of changes:
Add support for DBProxy CRD

Able to create DBProxy using yaml like this :

```
kind: DBProxy
metadata:
  name: "brucegu-proxy-0"
spec:
  dbProxyName: "brucegu-proxy-0"
  engineFamily: "POSTGRESQL"
  roleARN: "arn:aws:iam::316992738638:role/service-role/rds-proxy-role-1659373668348"
  auth:
  - secretARN: arn:aws:secretsmanager:us-west-2:316992738638:secret:prod/basepg-VCIT20
    authScheme: SECRETS
    iamAuth: DISABLED
  vpcSubnetIDs:
  - subnet-6d25e235
  - subnet-06a21f70
 ```
After proxy created, able to see it synced true
```
Status:
  Ack Resource Metadata:
    Arn:               arn:aws:rds:us-west-2:316992738638:db-proxy:prx-0ddcf5aa857718e04
    Owner Account ID:  316992738638
    Region:            us-west-2
  Conditions:
    Last Transition Time:  2022-08-01T22:02:39Z
    Message:               Resource synced successfully
    Reason:
    Status:                True
    Type:                  ACK.ResourceSynced
  Created Date:            2022-08-01T22:02:36Z
  Endpoint:                brucegu-proxy-0.proxy-c8jqtlwpfr20.us-west-2.rds.amazonaws.com
  Status:                  creating
  Updated Date:            2022-08-01T22:02:39Z
  Vpc ID:                  vpc-75137c11
Events:                    <none>
```

same applied to modify/delete operation. 

### CODE REVIEW

File: apis/v1alpha1/generator.yaml [hunk 1]
```diff
@@ -346,3 +346,12 @@ resources:
     fields:
       Name:
         is_primary_key: true
+  DBProxy:
+    exceptions:
+      terminal_codes:
+        - DBProxyAlreadyExistsFault
```
jaypipes, line 352: Hmm, not sure we need to list this one. This should only ever appear if the ACK runtime is seriously broken and the initial call to `ReadOne` for `DescribeDBProxies` returns NotFound by mistake and we try to create a DB proxy that already exists. I think it's actually safer to leave this one off the list of terminal conditions.

apis/v1alpha1/generator.yaml
```diff
@@ -346,3 +346,12 @@ resources:
...
+        - DBProxyQuotaExceededFault
+        - InvalidSubnet
+    fields:
+      DBProxyName:
```
jaypipes, line 356: Did you want to rename this to just `Name` to match the other RDS resources?

jaypipes (changes requested):
Not sure what happened with your local code-generator, but for some reason it's reset the Helm image tag to `v0.0.0-non-release-version`...

helm/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: rds-chart
 description: A Helm chart for the ACK service controller for Amazon Relational Database Service (RDS)
-version: v0.0.30
-appVersion: v0.0.30
+version: v0.0.0-non-release-version
+appVersion: v0.0.0-non-release-version
```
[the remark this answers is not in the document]
brucegucode: yeah, it's because i didn't do a git fetch for the latest tags, should be fixed in next revision.

apis/v1alpha1/generator.yaml
brucegucode: I think it's ok for now. previously we do the rename to avoid duplicate between spec and status, which is not an issue for DBProxy CRD here

generator.yaml
```diff
@@ -346,3 +346,10 @@ resources:
     fields:
       Name:
         is_primary_key: true
+  DBProxy:
+    exceptions:
+      terminal_codes:
+        - InvalidSubnet
+    fields:
+      DBProxyName:
```
jaypipes, line 354: did you want to rename this to `Name` like the other resources, @brucegucode?

brucegucode: Like i replied above, I think it's ok for now. previously we do the rename to avoid duplicate between spec and status, which is not an issue for DBProxy CRD here

jaypipes: Sorry, I missed your response above. That's not why we do the renames :) We do the renames to simplify and "de-stutter" the field names. For consistency with the existing resources like `DBParameterGroup`, `DBSubnetGroup` and `DBClusterParameterGroup` (the ones that are identified by a "name" field versus an "Identifier" field...) I think it would be best to rename this one here.

brucegucode: I agree with you to keep align with other resources

jaypipes (approved):
Cool, thanks Bruce! :)

PR #108, opened 2022-08-11
Comment out last multiAZ flag assert

brucegucode:
Issue #, if available:

Description of changes:
Comment out last multiAZ flag assertion

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2022-08-04T21:32:49Z"
+  build_date: "2022-08-13T17:21:53Z"
   build_hash: fe61d04673fd4d9848d5f726b01e0689a16d3733
   go_version: go1.18.1
   version: v0.19.3-1-gfe61d04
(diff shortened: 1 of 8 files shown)


jaypipes (approved):

brucegucode:
hmm, looks like it will fail when saz->maz flipped once, i will remove assert for second flip then

brucegucode:
Ah, it's the assert tag value, should match with controller version too

jaypipes:
@brucegucode yeah, we need to use the [`tags.assert_equal_without_ack_tags()`](https://github.com/aws-controllers-k8s/test-infra/blob/58ffb067d7c400ffbd0d470e975056efdc9e4965/src/acktest/tags.py#L119) function from the upstream test-infra package.

If not, we need to use the [`test.e2e.tags.clean`](https://github.com/aws-controllers-k8s/rds-controller/blob/f0864aecc24664b3ef8c3eed672addead0d6feaa/test/e2e/tag.py#L22) function.

jaypipes:
@brucegucode the reason the tag update isn't working is because [`compareTags` from `templates/hooks/db_instance/delta_pre_compare.go.tpl`](https://github.com/aws-controllers-k8s/rds-controller/blob/f0864aecc24664b3ef8c3eed672addead0d6feaa/templates/hooks/db_instance/delta_pre_compare.go.tpl#L13 ) is no longer being called. This is no longer being called because instead of using the template, the `delta_pre_compare` hook was [overridden with custom code](https://github.com/aws-controllers-k8s/rds-controller/blob/f0864aecc24664b3ef8c3eed672addead0d6feaa/generator.yaml#L203) in PR #95 while I was on vacation :)

PR #109, opened 2022-08-13
correct delta_pre_compare for DBInstance

jaypipes:
In PR #95, we added a `customPreCompare` function to
`pkg/resource/db_instance/hooks.go` and changed the `generator.yaml` to
use a code snippet instead of the
`templates/hooks/db_instance/delta_pre_compare.go.tpl` template file.
However, we did not delete the
`templates/hooks/db_instance/delta_pre_compare.go.tpl` file and when I
went to do the updating DB instance tags patch, nobody noticed that I
had added a call to `compareTags` into the (now-unused) template file.
This meant that `compareTags` is never called which means changes to
tags were not being properly on DBInstance resources.

This patch moves all the custom code for delta pre-checks back into the
template file and gets rid of the custom code snippet in generator.yaml,
consolidating all this logic into the template file.

Issue: aws-controllers-k8s/community#1276

Signed-off-by: Jay Pipes <jaypipes@gmail.com>

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.

### DIFF
--- a/test/e2e/tests/test_db_instance.py
+++ b/test/e2e/tests/test_db_instance.py
@@ -25,6 +25,7 @@ from e2e.replacement_values import REPLACEMENT_VALUES
 from e2e import condition
 from e2e import db_instance
 from e2e.fixtures import k8s_secret
+from e2e import tag
 
 RESOURCE_PLURAL = 'dbinstances'
 
@@ -180,7 +181,7 @@ class TestDBInstance:
         expect_tags = [
             {"Key": "environment", "Value": "dev"}
         ]
-        latest_tags = db_instance.get_tags(arn)
+        latest_tags = tag.clean(db_instance.get_tags(arn))
         assert expect_tags == latest_tags
 
         # OK, now let's update the tag set and check that the tags are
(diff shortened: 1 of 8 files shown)


A-Hilaly (approved):
:+1:

A-Hilaly:
e2e test failure is normal here, should be fixed in #108

A-Hilaly:
`gopkg.in` is struggling a lot lately..

test/e2e/tests/test_db_instance.py
```diff
@@ -197,7 +198,7 @@ def test_crud_postgres14_t3_micro(
         k8s.patch_custom_resource(ref, updates)
         time.sleep(MODIFY_WAIT_AFTER_SECONDS)
 
-        latest_tags = db_instance.get_tags(arn)
+        latest_tags = tag.clean(db_instance.get_tags(arn))
```
A-Hilaly, line 201: Good catch. Merging since the test failures are not accurate.
/lgtm

A-Hilaly:
Looks like gopkg.in/yaml.v2 is having some issues. Tracking the problem here https://github.com/niemeyer/gopkg/issues/78

PR #113, opened 2022-08-16
Update `computeTagsDelta` function

A-Hilaly:
This patch updates the `computeTagsDelta` function for `db_cluster`,
`db_parameter_group`, `db_subnet_group` and `db_cluster_parameter_group`
The previous implementation always returned a non-empty `toAdd` array
which causes the controller to make unnecessary `TagResource` API Calls.

Signed-off-by: Amine Hilaly <hilalyamine@gmail.com>


jkatz (commented):
Suggestion: given the code block is repeated across all the files, perhaps move the block to be an exportable function so the logic is in one place.

jkatz (approved):

jkatz (approved):

PR #115, opened 2022-08-22
Add e2e test for rdsproxy

brucegucode:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1440

Description of changes:

- Add e2e test for rds `dbproxy` resource

Two major issues were observed while testing locally

1. This test needs a complex dependency chain and it will cause bootstrap logic pretty complex. I'm inclined to either hard code it for our rds tests account or we just remove it for now.  The dependency chain is like this :  create one base rds instance -> create one aws secret service for this rds instance -> create an iam role for this secret -> create rds proxy using iam role & secret . Deletion is the opposite way.  Currently we're not able to create a rds instance in bootstrap logic or create a aws secret service secret. 
2. The e2e test lasts pretty long. creation will take around 10 min, deletion will take around 15 min, the total time would be around 30min in total which will greatly increase e2e test time. 

### DIFF
--- a/test/e2e/db_proxy.py
+++ b/test/e2e/db_proxy.py
@@ -1,0 +1,139 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""Utilities for working with DB proxy resources"""
+
+import datetime
+import time
+import typing
+
+import boto3
+import pytest
+
+DEFAULT_WAIT_UNTIL_TIMEOUT_SECONDS = 60*10
+DEFAULT_WAIT_UNTIL_INTERVAL_SECONDS = 15
+DEFAULT_WAIT_UNTIL_DELETED_TIMEOUT_SECONDS = 60*10
+DEFAULT_WAIT_UNTIL_DELETED_INTERVAL_SECONDS = 15
+
+ProxyMatchFunc = typing.NewType(
+    'ProxyMatchFunc',
+    typing.Callable[[dict], bool],
+)
+
+class StatusMatcher:
+    def __init__(self, status):
+        self.match_on = status
+
+    def __call__(self, record: dict) -> bool:
+        return 'Status' in record and record['Status'] == self.match_on
+
+
+def status_matches(status: str) -> ProxyMatchFunc:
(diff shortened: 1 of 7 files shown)


test/e2e/db_proxy.py
```diff
@@ -0,0 +1,139 @@
...
+    c = boto3.client('rds')
+    try:
+        resp = c.describe_db_proxies(DBProxyName=db_proxy_id)
+        assert len(resp['DBProxies']) == 1
+        return resp['DBProxies'][0]
+    except c.exceptions.DBProxyNotFoundFault:
+        return None
+
+
+def get_tags(db_proxy_arn):
```
A-Hilaly, line 127: Can you add tags supports and tags e2e tests for this resource too? just a copy-pasta from another resource's `hooks.go` should work

test/e2e/tests/test_db_proxy.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.mark.canary
+class TestDBProxy:
+
+    def test_crud_postgresql_proxy(
+            self,
+    ):
+        db_proxy_id = "my-test-proxy"
+        db_proxy_engine_family = "POSTGRESQL"
+        # The IAM role and secrect below has a complext dependency chain and we can hard code it for now
+        # It needs create one rds instance -> create aws secret based on it -> create IAM role based on this secret
```
A-Hilaly, line 50: Why is a `dbinstance` need to create a KMS secret?

test/e2e/tests/test_db_proxy.py
```diff
@@ -0,0 +1,100 @@
...
+        # I don't have a better way to fit this dependency chain in current rds controller yet
+        iam_role_arn = "arn:aws:iam::274006911594:role/Admin"
```
A-Hilaly, line 52: We have a bootstrappable resource to create iam_role_arns

A-Hilaly (commented):
Nice! Thanks @brucegucode. Left few comments inline

test/e2e/db_proxy.py
brucegucode: will try add it in next rev

test/e2e/tests/test_db_proxy.py
brucegucode: it's not a KMS secret, it's the AWS secret manager service's secret, hence need an actual rds instance to be up before creating a secret

test/e2e/tests/test_db_proxy.py
brucegucode: will try it in next rev

test/e2e/tests/test_db_proxy.py
jkatz: Let's add that as documentation.

A-Hilaly: Ohhh TIL!

brucegucode: I've made a comment in the above, updated a little bit in next rev

A-Hilaly (approved):
Thank you @brucegucode !
/lgtm

PR #123, opened 2023-01-07
Update AWS SDK to v1.44.174

jkatz:
This brings in the latest RDS API changes, including the ability to interface directly with Secrets Manager when creating a new DBInstance or DBCluster.

There is some follow-up work to do, i.e. determining how to integrate support for BlueGreenDeployment, but at a minimum this makes the RDS controller current with the existing APIs.


apis/v1alpha1/db_cluster.go
```diff
@@ -392,13 +394,54 @@ type DBClusterSpec struct {
...
+	// KMS key is used to encrypt the secret. If the secret is in a different Amazon
+	// Web Services account, then you can't use the aws/secretsmanager KMS key to
+	// encrypt the secret, and you must use a customer managed KMS key.
+	//
+	// There is a default KMS key for your Amazon Web Services account. Your Amazon
+	// Web Services account has a different default KMS key for each Amazon Web
+	// Services Region.
+	//
+	// Valid for: Aurora DB clusters and Multi-AZ DB clusters
+	MasterUserSecretKMSKeyID *string `json:"masterUserSecretKMSKeyID,omitempty"`
```
RedbackThomson, line 444: We can add a KMS key reference for this

apis/v1alpha1/db_cluster.go
```diff
@@ -815,6 +858,16 @@ type DBClusterStatus struct {
...
+	// Contains the secret managed by RDS in Amazon Web Services Secrets Manager
+	// for the master user password.
+	//
+	// For more information, see Password management with Amazon Web Services Secrets
+	// Manager (https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-secrets-manager.html)
+	// in the Amazon RDS User Guide and Password management with Amazon Web Services
+	// Secrets Manager (https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/rds-secrets-manager.html)
+	// in the Amazon Aurora User Guide.
+	// +kubebuilder:validation:Optional
+	MasterUserSecret *MasterUserSecret `json:"masterUserSecret,omitempty"`
```
RedbackThomson, line 870: I wonder if this will be confusing since the term `Secret` could refer to a K8s `Secret` too?

apis/v1alpha1/db_instance.go
```diff
@@ -629,6 +662,25 @@ type DBInstanceSpec struct {
...
+	//
+	// If you don't specify MasterUserSecretKmsKeyId, then the aws/secretsmanager
+	// KMS key is used to encrypt the secret. If the secret is in a different Amazon
+	// Web Services account, then you can't use the aws/secretsmanager KMS key to
+	// encrypt the secret, and you must use a customer managed KMS key.
+	//
+	// There is a default KMS key for your Amazon Web Services account. Your Amazon
+	// Web Services account has a different default KMS key for each Amazon Web
+	// Services Region.
+	MasterUserSecretKMSKeyID *string `json:"masterUserSecretKMSKeyID,omitempty"`
```
RedbackThomson, line 683: Also add ref for this

RedbackThomson (changes requested):
Just some basic questions and some new references. But otherwise good to go

apis/v1alpha1/db_cluster.go
jkatz: Agreed. OTOH this is the name used in the upstream API, so someone referencing the RDS API docs would look for this. Need to think on this.

RedbackThomson: I think the Go docs (which will make its way into the API reference on the website) do a pretty good of explaining this

A-Hilaly:
I think the new code-generator version is breaking the compilation. Maybe we should try with v0.20.0?

vmercierfr:
Thanks for upgrading the AWS SDK version. It will help us to deploy instances with GP3 storage type.
However, we may need to add a small logic to support this new storage type correctly.

The GP3 storage type has a specific logic regarding IOPS depending on the DB engine and the size of the storage size:
 
> For every DB engine except RDS for SQL Server, you can provision additional IOPS and storage throughput when storage size is at or above the threshold value
> https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#gp3-storage

TL;DR If storage size is below 400 GB, you can't specify IOPS for _MariaDB, MySQL, PostgreSQL, and Oracle_ DB engines.

Currently, if you call the `ModifyDBInstance` RDS API method with `allocatedStorage < 400` and specify `iops` parameter, it will return an `InvalidParameterCombinationException` error with `You can't specify IOPS or storage throughput for engine <engine> and a storage size less than 400` error message.

Seems all projects implementing `gp3` storage for RDS have to deal with this behavior and avoid specifying `iops` parameter for storage < 400 GB when calling the RDS API. For instance, Terraform which is also using AWS SDK had a bug (https://github.com/hashicorp/terraform-provider-aws/issues/28271) and fix it by adding [the threshold logic in the code](https://github.com/hashicorp/terraform-provider-aws/pull/28243/files#diff-5684e38aee8a02cd8ffe882cacb6fccda55107f3f73d76a5aec501bb07713035R1893-R1896) 

To test it :

1. Create an RDS instance with `storageType=gp3` and `storageSize=20`
2. Try to change an RDS parameter (eg. storage size, engine version, ...)
3. API RDS should return an InvalidParameterCombinationException error

Can you determine if the "400 GB threshold" must be added to the `rds-controller`?

jaypipes:
Closing in favor of #144

PR #133, opened 2023-03-07
Allow parameter overrides for a DB cluster parameter group

jljaco:
Issue #, if available:
https://github.com/aws-controllers-k8s/community/issues/869

The `Spec.ParameterOverrides` is a `map[string]*string` containing the parameter names and values for parameters in a DB cluster parameter group for which the user wishes to override the defaults.

The `Status.ParameterOverrideStatuses` field is a slice of `Parameter` structs that contains status information about those overridden parameters.


jljaco:
This is still a work in progress

config/crd/bases/rds.services.k8s.aws_dbclusterparametergroups.yaml
```diff
@@ -192,6 +153,41 @@ spec:
                   - type
                   type: object
                 type: array
+              parameterOverrideStatuses:
+                description: Provides a list of parameters for the DB cluster parameter
+                  group.
+                items:
+                  description: "This data type is used as a request parameter in the
```
jljaco, line 160: This misleading comment extracted automatically from https://github.com/aws/aws-sdk-go/blob/6cdd6178046217238d1969f5343fd929d7173703/service/rds/api.go#L44729

helm/crds/rds.services.k8s.aws_dbclusterparametergroups.yaml
```diff
@@ -193,6 +154,41 @@ spec:
                   - type
                   type: object
                 type: array
+              parameterOverrideStatuses:
+                description: Provides a list of parameters for the DB cluster parameter
+                  group.
+                items:
+                  description: "This data type is used as a request parameter in the
```
jljaco, line 161: Ditto -- automatically extracted from https://github.com/aws/aws-sdk-go/blob/6cdd6178046217238d1969f5343fd929d7173703/service/rds/api.go#L44729

pkg/resource/db_cluster_parameter_group/delta.go
```diff
@@ -63,8 +63,12 @@ func newResourceDelta(
 			delta.Add("Spec.Name", a.ko.Spec.Name, b.ko.Spec.Name)
 		}
 	}
-	if !reflect.DeepEqual(a.ko.Spec.Parameters, b.ko.Spec.Parameters) {
-		delta.Add("Spec.Parameters", a.ko.Spec.Parameters, b.ko.Spec.Parameters)
+	if ackcompare.HasNilDifference(a.ko.Spec.ParameterOverrides, b.ko.Spec.ParameterOverrides) {
```
jljaco, line 66: Noting this comment from the analogous line in #107:

https://github.com/aws-controllers-k8s/rds-controller/pull/107/files#r948248230

jljaco:
Noting that the e2e tests are currently [failing](https://prow.ack.aws.dev/view/s3/ack-prow-logs/pr-logs/pull/aws-controllers-k8s_rds-controller/133/rds-kind-e2e/1633544817986244608) for an unrelated reason.

The assertion [here](https://github.com/aws-controllers-k8s/rds-controller/blob/c288c4f1e60fd5d9c86ef294690bd87da4bf007c/test/e2e/tests/test_db_proxy.py#L79) is failing.

I am specifically not going to fix that problem in this PR; it should be fixed in another one.

apis/v1alpha1/db_cluster_parameter_group.go
```diff
@@ -88,20 +88,8 @@ type DBClusterParameterGroupSpec struct {
...
-	// the pending-reboot value for both dynamic and static parameters.
-	//
-	// When the application method is immediate, changes to dynamic parameters are
-	// applied immediately to the DB clusters associated with the parameter group.
-	// When the application method is pending-reboot, changes to dynamic and static
-	// parameters are applied after a reboot without failover to the DB clusters
-	// associated with the parameter group.
-	Parameters []*Parameter `json:"parameters,omitempty"`
+	Name               *string            `json:"name"`
+	ParameterOverrides map[string]*string `json:"parameterOverrides,omitempty"`
```
RedbackThomson, line 92: We 100% cannot remove, or modify the shape of, an existing spec field. It breaks backwards compatibility and would have pretty strange effects for existing customers. I understand this complicates the implementation, but if we can fit the changes within the existing spec that would be nice, otherwise we have to add a new field altogether.

pkg/resource/db_cluster_parameter_group/hooks.go
```diff
@@ -173,3 +196,234 @@ func sdkTagsFromResourceTags(
...
+	err error,
+) {
+	var marker *string
+	params = make(map[string]*string)
+	for {
+		resp, err := rm.sdkapi.DescribeDBClusterParametersWithContext(
+			ctx,
+			&svcsdk.DescribeDBClusterParametersInput{
+				DBClusterParameterGroupName: groupName,
+				Source:                      aws.String(sourceUser),
```
RedbackThomson, line 261: I have read the RDS docs but I still don't understand the significance of this. Is this relevant to our controller at all?

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+	// from their list of parameter overrides, we will not be able to get the
+	// resource into a synced state.
+	return ackerr.NewTerminalError(
+		fmt.Errorf("%w: %s", errUnmodifiableParameter, name),
+	)
+}
+
+// ComputeParametersDelta compares two Parameter arrays and returns the new
+// parameters to add, to update and the parameter identifiers to delete
+func ComputeParametersDelta(
```
RedbackThomson, line 36: Check out how I did this for tags in the runtime using `lo` helper methods - https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/compare/tags.go#L26

It might not be any shorter, but I think the code is more maintainable

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+		if _, found := desired[k]; !found {
+			toReset = append(toReset, k)
+		}
+	}
+	return toModify, toReset
+}
+
+// sliceStringChunks splits a supplied slice of string pointers into multiple
+// slices of string pointers of a given size.
+func SliceStringChunks(
```
RedbackThomson, line 60: `samber/lo` supports this out of the box - https://github.com/samber/lo#chunk

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+			chunk = map[string]*string{}
+			idx = 0
+		}
+	}
+	chunks = append(chunks, chunk)
+
+	return chunks
+}
+
+// *** Caching
```
RedbackThomson, line 106: nit: move all of this under a new file `caching.go`

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
+package util
```
RedbackThomson, line 1: This file needs an Amazon license header

RedbackThomson (changes requested):
Really excellent and clean code with the custom hook refactoring. Unfortunately there is a major issue with modifying the existing API shapes and/or names, that I've commented on below.

apis/v1alpha1/db_cluster_parameter_group.go
jljaco: OK, just to be clear, is this only the rule for GA controllers?

Because an analogous change was made last September for `db_parameter_group` [here](https://github.com/aws-controllers-k8s/rds-controller/pull/107/files#diff-fbd5c921ed4d5f67af22264e781b19e636ba01ca7e8c8c95c043bfb8ccb3ecb2)

A-Hilaly: >OK, just to be clear, is this only the rule for GA controllers?

Yes.

>Because an analogous change was made last September for db_parameter_group [here](https://github.com/aws-controllers-k8s/rds-controller/pull/107/files#diff-fbd5c921ed4d5f67af22264e781b19e636ba01ca7e8c8c95c043bfb8ccb3ecb2)

That change was made few days before going GA :/. If we don't have multi-version support we cannot delete/rename any fields.

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
+)
+
+var (
+	errUnknownParameter      = fmt.Errorf("unknown parameter")
+	errUnmodifiableParameter = fmt.Errorf("parameter is not modifiable")
```
A-Hilaly, line 13: nit: i know those errors are not imported anywhere, but since they are pre-defined errors they need to be exposed (public) in case in one wants to select types on them.

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+func NewErrUnknownParameter(name string) error {
+	// This is a terminal error because unless the user removes this parameter
+	// from their list of parameter overrides, we will not be able to get the
+	// resource into a synced state.
+	return ackerr.NewTerminalError(
+		fmt.Errorf("%w: %s", errUnknownParameter, name),
+	)
+}
+
+func NewErrUnmodifiableParameter(name string) error {
```
A-Hilaly, line 25: Public functions/objects needs to have a comment - https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	for k := range latest {
+		if _, found := desired[k]; !found {
+			toReset = append(toReset, k)
+		}
+	}
+	return toModify, toReset
+}
+
+// sliceStringChunks splits a supplied slice of string pointers into multiple
```
A-Hilaly, line 58:
```suggestion
// SliceStringChunks splits a supplied slice of string pointers into multiple
```

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		chunks = append(chunks, input[0:chunkSize])
+		input = input[chunkSize:]
+	}
+
+	return chunks
+}
+
+// mapStringChunks splits a supplied map of string pointers into multiple
```
A-Hilaly, line 81:
```suggestion
// MapStringChunks splits a supplied map of string pointers into multiple
```

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	chunks = append(chunks, chunk)
+
+	return chunks
+}
+
+// *** Caching
+
+type ParamMeta struct {
```
A-Hilaly, line 108: ditto

pkg/util/parameters.go
```diff
@@ -0,0 +1,180 @@
...
+	IsModifiable bool
+	IsDynamic    bool
+}
+
+// metaFetcher is the functor we pass to the paramMetaCache that allows it to
+// fetch engine default parameter information
+type metaFetcher func(ctx context.Context, family string) (map[string]ParamMeta, error)
```
A-Hilaly, line 115: Preferably `MetaFetcher` needs to be exported

A-Hilaly (commented):
Good stuff! left few tiny suggestions.

pkg/resource/db_cluster_parameter_group/hooks.go
jaypipes: @RedbackThomson there's no need to return any non-user-settable parameters, which is why this filter is passed.

pkg/util/parameters.go
jaypipes: Or `param_cache.go`

jaypipes (commented):
It's a shame the RDS controller was GA'd before this could be done. I almost prefer to make the backwards incompatible CRD schema changes so that this aligns with the DBParameterGroup CRD... especially since we're still at v1alpha1 API version...

apis/v1alpha1/db_cluster_parameter_group.go
jljaco: As discussed in the community meeting this morning, it is worth noting that this is a field that hasn't actually ever worked.  I'm a bit dubious that anyone might actually be using it.  Also, there is a value to having `db_parameter_group` and `db_cluster_parameter_group` behave similarly, since they're both effectively doing the same thing.

pkg/util/parameters.go
jljaco: Done

pkg/util/parameters.go
jljaco: Done

(`parameter_cache.go`)

pkg/util/parameters.go
jljaco: Done

pkg/util/parameters.go
jljaco: Done

pkg/util/parameters.go
jljaco: Done

pkg/util/parameters.go
jljaco: Done

pkg/util/parameters.go
jljaco: Done

pkg/util/parameters.go
jljaco: Done

jaypipes (changes requested):
@jljaco I believe you still need to 1) add a call to `syncParameters` somewhere (I don't see either `customUpdate` changed nor do I see any template/hook code call `syncParameters` so parameter overrides will never be set. 2) modify `test/e2e/tests/test_db_cluster_parameter_group.py` to add the parameter override testing

apis/v1alpha1/db_cluster_parameter_group.go
jljaco: Restored `Parameters` in `generator.yaml` and marked it as `DEPRECATED` using custom doc strings

go.local.mod
```diff
@@ -50,10 +50,10 @@ require (
...
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/prometheus/client_golang v1.14.0 // indirect
-	github.com/prometheus/client_model v0.3.0 // indirect
-	github.com/prometheus/common v0.37.0 // indirect
-	github.com/prometheus/procfs v0.8.0 // indirect
+	github.com/prometheus/client_golang v1.11.0 // indirect
+	github.com/prometheus/client_model v0.2.0 // indirect
+	github.com/prometheus/common v0.28.0 // indirect
+	github.com/prometheus/procfs v0.6.0 // indirect
```
A-Hilaly, line 56: Maybe a rebase issue? client_golang should stay `v1.14.0`. Source: https://github.com/aws-controllers-k8s/runtime/blob/main/go.mod#L13

A-Hilaly: Mmmmm looks like this the local one... please discard my comment :)

pkg/util/parameters.go
jljaco: Done

templates/hooks/db_cluster_parameter_group/sdk_read_many_post_set_output.go.tpl
```diff
@@ -1,8 +1,17 @@
...
+    }
+    if ko.Spec.Name != nil {
+        groupName := ko.Spec.Name
+        params, paramStatuses, err := rm.getParameters(ctx, groupName)
+        if err != nil {
+            return nil, err
+        }
+        ko.Spec.ParameterOverrides = params
+        ko.Status.ParameterOverrideStatuses = paramStatuses
+    }
```
jaypipes, line 17: Not necessary to fix in this PR, but just a quick tip... you should replace four spaces with `\t` characters in all the templates. Likely your IDE or text editor just isn't using TABs when the code file isn't a `.go` file...

jaypipes (approved):
I think this is ready to go. Thank you @jljaco :)

pkg/util/parameters.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+var (
+	ErrUnknownParameter      = fmt.Errorf("unknown parameter")
+	ErrUnmodifiableParameter = fmt.Errorf("parameter is not modifiable")
+)
+
+// Parameters represents the elements of a DB Parameter Group
+// or a DB Cluster Parameter Group
+type Parameters map[string]*string
```
A-Hilaly, line 30: Type aliasing for basic Go type is very rarely recommended in idiomatic Go - i'd suggest just use `map[string]*string` in the function parameters

pkg/util/parameters.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return added, unchanged, removed
+}
+
+// ChunkMap splits a supplied map of parameters into multiple
+// slices of maps of parameters of a given size.
+func ChunkMap(
+	input Parameters,
+	chunkSize int,
+) []Parameters {
```
A-Hilaly, line 79: `ChunkMap` is good name if input was `map[string]*string`, otherwise it should be `ChunkParameters`

pkg/resource/db_cluster_parameter_group/hooks.go
```diff
@@ -15,21 +15,44 @@ package db_cluster_parameter_group
...
 
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 
 	svcapitypes "github.com/aws-controllers-k8s/rds-controller/apis/v1alpha1"
+	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/rds"
 
 	"github.com/aws-controllers-k8s/rds-controller/pkg/util"
```
A-Hilaly, line 29: https://github.com/golang/go/wiki/CodeReviewComments#imports

pkg/util/parameter_cache.go
```diff
@@ -0,0 +1,94 @@
...
+// family. We use this cached information to determine whether a parameter is
+// statically or dynamically defined (whether changes can be applied
+// immediately or pending a reboot) and whether a parameter is modifiable.
+//
+// Keeping things super simple for now and not adding any TTL or expiration
+// behaviour to the cache. Engine defaults are pretty static information...
+type ParamMetaCache struct {
+	sync.RWMutex
+	Hits   uint64
+	Misses uint64
```
A-Hilaly, line 41: @jaypipes I still don't understand the use of Hits and Missed in this implementation, they are nor used nor logged, just incremented - did you have other further intentions for this?

test/e2e/tests/test_db_cluster_parameter_group.py
```diff
@@ -82,8 +83,14 @@ def test_create_delete_aurora_mysql5_7(self):
...
         ]
+        new_params = {
+            "autocommit": "1",
+        }
         updates = {
-            "spec": {"tags": new_tags},
+            "spec": {
+                "tags": new_tags,
+                "parameterOverrides": new_params,
+            },
```
A-Hilaly, line 93: Since this implementation is exceptionally complex - i'd suggest testing more than just one parameter.

A-Hilaly (changes requested):
Good job on this! left few comments inline

pkg/util/parameters.go
jljaco: https://github.com/aws-controllers-k8s/runtime/blob/fb553741e0bfbbd1fee8f52b2e36b8c58dd7bf8e/pkg/tags/tags.go#L21

My intent was to create an analog to the [GetTagsDifference](https://github.com/aws-controllers-k8s/runtime/blob/7152103f915b2722bef679ebe6df231e7f489222/pkg/compare/tags.go#L26) function

as [suggested here](https://github.com/aws-controllers-k8s/rds-controller/pull/133#discussion_r1130237940)

pkg/util/parameters.go
jljaco: Done

pkg/resource/db_cluster_parameter_group/hooks.go
```diff
@@ -173,3 +201,235 @@ func sdkTagsFromResourceTags(
 	}
 	return tags
 }
+
+func equalStrings(a, b *string) bool {
+	if a == nil {
+		return b == nil || *b == ""
+	}
+	return (*a == "" && b == nil) || *a == *b
+}
```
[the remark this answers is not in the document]
jljaco: Noted.  As you've pointed out in that comment, the issue applies to many different instances of this function across the codebase.  Would you prefer that I fix this one instance, knowing that it will be out of alignment with all of the other, broken instances?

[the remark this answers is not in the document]
A-Hilaly: Ideally we should fix it now. Let's not wait until it lands in runtime. Updating all the controllers will take alot of time.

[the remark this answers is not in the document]
jljaco: Turns out that this function was unused, so I just pruned it.

pkg/util/parameter_cache.go
```diff
@@ -0,0 +1,94 @@
...
+	name string,
+	fetcher MetaFetcher,
+) (*ParamMeta, error) {
+	var err error
+	var found bool
+	var metas map[string]ParamMeta
+	var meta ParamMeta
+
+	// We need to release the lock right after the read operation, because
+	// loadFamilly will might call a writeLock at L619
```
RedbackThomson, line 59: nit: Update the line number in this comment to match new file location

RedbackThomson (approved):
Really really excellent work.
If I were to nitpick the testing, I might say that we should test adding params in the initial create - so we can test that hook code you've added for requeueing instantly after create. Also possibly adding another patch to remove a param, so we know resetting parameters works.

pkg/util/parameter_cache.go
jljaco: Removed the line number entirely

test/e2e/tests/test_db_cluster_parameter_group.py
```diff
@@ -97,15 +97,18 @@ def test_create_delete_aurora_mysql5_7(
...
+                found += 1
                 assert "ParameterValue" in tp, f"No ParameterValue in parameter of name 'aurora_binlog_read_buffer_size': {tp}"
                 assert tp["ParameterValue"] == "8192", f"Wrong value for parameter of name 'aurora_binlog_read_buffer_size': {tp}"
-                break
-        assert found, f"No parameter of name 'aurora_binlog_read_buffer_size' was found: {test_params}"
+            elif tp["ParameterName"] == "aurora_read_replica_read_committed":
+                found += 1
+                assert "ParameterValue" in tp, f"No ParameterValue in parameter of name 'aurora_read_replica_read_committed': {tp}"
+                assert tp["ParameterValue"] == "OFF", f"Wrong value for parameter of name 'aurora_read_replica_read_committed': {tp}"
+        assert found == 2, f"Did not find parameters with names 'aurora_binlog_read_buffer_size' and 'aurora_read_replica_read_committed': {test_params}"
```
A-Hilaly, line 111: QQ: I guess you're iterating on the array (and using `found`var) because the order of `tp` is not deterministic?

A-Hilaly (approved):
Looks good and clean! Thank you @jljaco

james-callahan:
It appears this may have been reverted. The CRD comment particularly in https://github.com/aws-controllers-k8s/rds-controller/pull/155/files#diff-2b8488e2b84c0c1089daa216c5df7bd1caadee6e2af5e378590957856bae5560L107

PR #135, opened 2023-03-08
Copy the security groups from status into spec

arjunrn:
Issue #, if available: aws-controllers-k8s/community/issues/1712

Description of changes:
Because the names of the security group fields returned by the API differ the `VPCSecurityGroupsIDs` is not read from the API response into the _latest_. Instead the values of the _desired_ is copied. This means that the controller determines that there is not need for an update. Instead by copying the the VPCSecurityGroups field from the status into the spec the controller can correctly determine that there is a difference between the latest and desired. This affects both type __DBCluster__ and __DBInstance__.


jaypipes (approved):
@arjunrn thank you for getting to this so quickly! Looks great to me.

jaypipes:
@A-Hilaly can we merge this one? test fails are unrelated...

A-Hilaly:
@jaypipes yep merging..

PR #137, opened 2023-03-10
Add asynchronous status checks for `DBProxy`

RedbackThomson:
Description of changes:
The DBProxy APIs put the resource into `creating` state for ~10 minutes before becoming `available`, as well as `deleting` before it is gone. This PR copies the systems we use for `DBInstance`, which has the same properties, into custom hooks for `DBProxy`. It also updates the tests for `DBProxy` to use proper fixturing and to use a autogenerated resource name suffix (so tests don't collide).

### DIFF
--- a/pkg/resource/db_proxy/hooks.go
+++ b/pkg/resource/db_proxy/hooks.go
@@ -15,8 +15,11 @@ package db_proxy
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	ackrequeue "github.com/aws-controllers-k8s/runtime/pkg/requeue"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 
 	svcapitypes "github.com/aws-controllers-k8s/rds-controller/apis/v1alpha1"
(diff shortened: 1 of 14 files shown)


jljaco (approved):
Looks fine to me, will leave it to @A-Hilaly for final approval

A-Hilaly (approved):
This looks great! :+1:

pkg/resource/db_proxy/hooks.go
```diff
@@ -25,6 +28,86 @@ import (
 	"github.com/aws-controllers-k8s/rds-controller/pkg/util"
 )
 
+var (
+	// TerminalStatuses are the status strings that are terminal states for a
+	// DB proxy.
+	TerminalStatuses = []string{
+		svcsdk.DBProxyStatusIncompatibleNetwork,
+		svcsdk.DBProxyStatusDeleting,
+		svcsdk.DBProxyStatusSuspending,
```
jaypipes, line 37: Are these terminal statuses though? These aren't statuses that would require the user to change something in the `Spec` so probably should not be terminal conditions...

RedbackThomson: `Deleting` is a terminal status for `DBInstance` - https://github.com/aws-controllers-k8s/rds-controller/blob/main/pkg/resource/db_instance/hooks.go#L89

As for `Suspending`, I figured it was inline with `Deleting` as a status that users could not get out from by modifying spec.

jaypipes: Hmmm, yeah. I guess it's OK since this will only cause a Terminal condition to appear on the resource IFF the user has deleted or suspended a DBProxy AND THEN made a change to a Spec field, since this is only called on sdkUpdate

jaypipes (approved):
:+1:

A-Hilaly:
Merging as is since test failures are not relevant (#133 is testing in parallel)

PR #148, opened 2023-05-01
Update template to support watch namespace

rahtr:
As the controller runtime supports watching namespaces, we would need to add a support to watch namespaces other than where the controller is deployed. This is a common use case for multi-tenant K8s environments. This PR introduces a `watchNamespace` flag which lets user define the namespaces to be watched by the controller. If left blank, it would take the release namespace.

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -33,7 +33,7 @@ If release name contains chart name it will be used as a full name.
 
 {{- define "watch-namespace" -}}
 {{- if eq .Values.installScope "namespace" -}}
-{{- .Release.Namespace -}}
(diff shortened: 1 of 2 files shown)


at-byte:
We would like to see this support as well.

jaypipes (commented):
Hi @rahtr thank you for your PR. All of the ACK's controllers are generated, so you will need to make this change against the Helm template files in the ACK code generator, here:

https://github.com/aws-controllers-k8s/code-generator/tree/main/templates/helm/templates

That said, I'm not sure where you are getting the `watchNamespace` value, since there is no such value in our [values.yaml](https://github.com/aws-controllers-k8s/code-generator/blob/main/templates/helm/values.yaml.tpl) file.

PR #151, opened 2023-05-11
prevent DB Instance desired field state flapping

jaypipes:
Both ModifyDBInstance and DescribeDBInstances returns a DBInstance struct that contains the *previously set* values for various mutable fields. This is problematic because it causes a "flopping" behaviour when the user has modified a Spec field from value A to value B but the output shape from ModifyDBInstance for that field contains value A, the standard SetOutput Go code generated above will set the Spec field to the *old* value again. The next time the reconciler runs, it will read the latest observed resource, see a difference between the desired and the latest state (that actually does not exist because the difference is comparing the value of the fields before they were changed) and attempt to modify the field from value B to value A again, causing a flop loop.

Luckily, the Output shape's DBInstance struct contains a `PendingModifiedValues` struct which contains those field values that the user specified. So, we can use these to "reset" the Spec back to the appropriate user-specified values.

This commit does exactly this. It looks in the `PendingModifiedValues` struct for any non-nil field that matches a field in the `Spec` and if it finds a match, sets the `Spec` field to the value in `PendingModifiedValues`. This solves the flapping problem for all of the fields in `PendingModifiedValues`, including `DBInstanceClass`, `AllocatedStorage`, `MultiAZ` and `StorageType`.

Fixes Issue aws-controllers-k8s/community#1773
Fixes Issue aws-controllers-k8s/community#1716
Fixes Issue aws-controllers-k8s/community#1376

### DIFF
--- a/templates/hooks/db_instance/sdk_read_many_post_set_output.go.tpl
+++ b/templates/hooks/db_instance/sdk_read_many_post_set_output.go.tpl
@@ -1,10 +1,94 @@
+	// DescribeDBInstances returns an array of DBInstance structs that contains
+	// the *previously set* values for various mutable fields. This is
+	// problematic because it causes a "flopping" behaviour when the user has
+	// modified a Spec field from value A to value B but the output shape from
+	// ModifyDBInstance for that field contains value A, the standard SetOutput
+	// Go code generated above will set the Spec field to the *old* value
+	// again. The next time the reconciler runs, it will read the latest
+	// observed resource, see a difference between the desired and the latest
+	// state (that actually does not exist because the difference is comparing
+	// the value of the fields before they were changed) and attempt to modify
+	// the field from value B to value A again, causing a flop loop.
+	//
+	// Luckily, the Output shape's DBInstance struct contains a
+	// `PendingModifiedValues` struct which contains those field values that
+	// the user specified. So, we can use these to "reset" the Spec back to the
+	// appropriate user-specified values.
+	pmv := ko.Status.PendingModifiedValues
+	if pmv != nil {
+		if pmv.AllocatedStorage != nil {
+			ko.Spec.AllocatedStorage = pmv.AllocatedStorage
+		}
+        // NOTE(jaypipes): Handle when aws-sdk-go update
+		//if pmv.AutomationMode != nil {
+		//	ko.Spec.AutomationMode = pmv.AutomationMode
+		//}
+		if pmv.BackupRetentionPeriod != nil {
+			ko.Spec.BackupRetentionPeriod = pmv.BackupRetentionPeriod
+		}
+		if pmv.CACertificateIdentifier != nil {
+			ko.Spec.CACertificateIdentifier = pmv.CACertificateIdentifier
+		}
+		if pmv.DBInstanceClass != nil {
+			ko.Spec.DBInstanceClass = pmv.DBInstanceClass
+		}
+		if pmv.DBInstanceIdentifier != nil {
+			ko.Spec.DBInstanceIdentifier = pmv.DBInstanceIdentifier
+		}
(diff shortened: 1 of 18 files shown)


templates/hooks/db_instance/sdk_read_many_post_set_output.go.tpl
```diff
@@ -1,10 +1,88 @@
...
+	// Luckily, the Output shape's DBInstance struct contains a
+	// `PendingModifiedValues` struct which contains those field values that
+	// the user specified. So, we can use these to "reset" the Spec back to the
+	// appropriate user-specified values.
+	pmv := ko.Status.PendingModifiedValues
+	if pmv != nil {
+		if pmv.AllocatedStorage != nil {
+			ko.Spec.AllocatedStorage = pmv.AllocatedStorage
+		}
+		//if pmv.AutomationMode != nil {
```
jljaco, line 22: What's the logic behind skipping `AutomationMode` (as well as other fields, below)?

Could you add a comment here explaining why it isn't necessary to also make this copy?

templates/hooks/db_instance/sdk_update_post_set_output.go.tpl
```diff
@@ -1,8 +1,83 @@
...
+	// Luckily, the Output shape's DBInstance struct contains a
+	// `PendingModifiedValues` struct which contains those field values that
+	// the user specified. So, we can use these to "reset" the Spec back to the
+	// appropriate user-specific values.
+	pmv := resp.DBInstance.PendingModifiedValues
+	if pmv != nil {
+		if pmv.AllocatedStorage != nil {
+			ko.Spec.AllocatedStorage = pmv.AllocatedStorage
+		}
+		//if pmv.AutomationMode != nil {
```
jljaco, line 19: Same question/request about adding a comment explaining why we're skipping these fields.

test/e2e/tests/test_db_instance.py
```diff
@@ -97,14 +97,14 @@ def postgres14_t3_micro_instance(k8s_secret):
...
-        db_instance.wait_until_deleted(db_instance_id)
     except:
         pass
+    db_instance.wait_until_deleted(db_instance_id)
+
 
 @service_marker
 @pytest.mark.canary
-class TestDBInstance:  
+class TestDBInstance:
```
A-Hilaly, line 107: Can we uncomment the tests below in `L164` and `L180`?

templates/hooks/db_instance/sdk_read_many_post_set_output.go.tpl
jaypipes: @jljaco those fields are included in later versions of aws-sdk-go. I will uncomment them in a followup PR when I update the aws-sdk-go in go.mod to a more recent version. I wanted to try and reduce the amount of changes in this PR and left these commented out to remind me to handle them for the newer aws-sdk-go version fields.

test/e2e/tests/test_db_instance.py
jaypipes: > Can we uncomment the tests below in `L164` and `L180`?

good catch. yeah. I will update.

templates/hooks/db_instance/sdk_read_many_post_set_output.go.tpl
jljaco: Cool, understood!  Could you add a comment in the code to that effect?

templates/hooks/db_instance/sdk_update_post_set_output.go.tpl
jaypipes: @jljaco all done.

A-Hilaly (approved):
Mucho gracias @jaypipes !!
/lgtm

PR #159, opened 2023-11-07
Copy IAM authentication state from status

arjunrn:
Issue #, if available: #1937

Description of changes: Like the parameter groups the IAM authentication flag has different names in the input and output to the API. A hook is need to correctly set the value.

### DIFF
--- a/test/e2e/tests/test_db_cluster.py
+++ b/test/e2e/tests/test_db_cluster.py
@@ -65,7 +65,7 @@ def aurora_mysql_cluster(k8s_secret):
     db_name = "mydb"
     secret = k8s_secret(
         MUP_NS,
-        MUP_SEC_NAME,
+        f"{MUP_SEC_NAME}-mysql",
         MUP_SEC_KEY,
         MUP_SEC_VAL,
     )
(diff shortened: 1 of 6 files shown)


test/e2e/tests/test_db_cluster.py
```diff
@@ -108,6 +108,53 @@ def aurora_mysql_cluster(k8s_secret):
...
+
+@pytest.fixture
+def aurora_postgres_cluster(k8s_secret):
+    db_cluster_id = random_suffix_name("my-aurora-postgres", 20)
+    secret = k8s_secret(
+        MUP_NS,
+        MUP_SEC_NAME,
+        MUP_SEC_KEY,
+        MUP_SEC_VAL,
+    )
```
a-hilaly, line 120: Thanks @arjunrn ! We should use a different name here to avoid conflict with other created secrets

arjunrn: I've fixed the naming. PTAL.

a-hilaly (approved):
Thanks @arjunrn !
/lgtm

PR #166, opened 2024-02-09
Ignore EngineVersion minor updates when `autoMinorVersionUpgrade` is enabled

joaquinservetto:
Fixes [Issue](https://github.com/aws-controllers-k8s/community/issues/2009)

### DIFF
--- a/pkg/resource/db_cluster/custom_update.go
+++ b/pkg/resource/db_cluster/custom_update.go
@@ -16,6 +16,7 @@ package db_cluster
 import (
 	"context"
 	"slices"
+	"regexp"
 
 	ackv1alpha1 "github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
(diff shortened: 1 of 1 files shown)


pkg/resource/db_cluster/custom_update.go
```diff
@@ -561,7 +562,12 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 		res.SetEnableIAMDatabaseAuthentication(*desired.ko.Spec.EnableIAMDatabaseAuthentication)
 	}
 	if desired.ko.Spec.EngineVersion != nil && delta.DifferentAt("Spec.EngineVersion") {
-		res.SetEngineVersion(*desired.ko.Spec.EngineVersion)
+		r := regexp.MustCompile(`[0-9]*$`)
```
a-hilaly, line 565: Can we delcare this regex as a total level file variable? just avoid compiling the regex at every reconciliation loop

pkg/resource/db_cluster/custom_update.go
```diff
@@ -561,7 +562,12 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
...
+		desiredMajorEngineVersion := r.ReplaceAllString(*desired.ko.Spec.EngineVersion, "${1}")
```
a-hilaly, line 566: nil, we need to handle the case of nil `*desired.ko.Spec.EngineVersion`

pkg/resource/db_cluster/custom_update.go
```diff
@@ -561,7 +562,12 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 		res.SetEnableIAMDatabaseAuthentication(*desired.ko.Spec.EnableIAMDatabaseAuthentication)
 	}
 	if desired.ko.Spec.EngineVersion != nil && delta.DifferentAt("Spec.EngineVersion") {
```
a-hilaly, line 564: Can we move this logic into a little function called `requireEngineVersionUpdate` or something similar? that was we could add unit tests for it :)

a-hilaly (commented):
This is great start, muchas gracias @joaquinservetto ! i left few recommendations below
/ok-to-test

pkg/resource/db_cluster/custom_update.go
joaquinservetto: Thanks a lot for you review @a-hilaly. If `*desired.ko.Spec.EngineVersion` is nil it'll not perform any update, isn't it the expected behavior?

a-hilaly (approved):
Thank you @joaquinservetto !
/lgtm

PR #182, opened 2024-04-09
Generate `DBClusterSnapshot` resource

a-hilaly:
This commit adds support for `DBClusterSnapshot` resources. Users can now create,
delete and update `DBClusterSnapshots` for RDS cluster snapshots using the ACK
RDS controller.

- Implement resource manager for DBClusterSnapshot with Create, Delete,
  ReadOne and Update SDK calls
- Add custom Update logic to sync tags
- Check cluster snapshot status and requeue if cluster snapshot is not in
  'available' state
- Adds e2e tests


pkg/resource/db_cluster_snapshot/hooks.go
```diff
@@ -0,0 +1,220 @@
...
+	b *resource,
+) {
+	if len(a.ko.Spec.Tags) != len(b.ko.Spec.Tags) {
+		delta.Add("Spec.Tags", a.ko.Spec.Tags, b.ko.Spec.Tags)
+	} else if len(a.ko.Spec.Tags) > 0 {
+		if !util.EqualTags(a.ko.Spec.Tags, b.ko.Spec.Tags) {
+			delta.Add("Spec.Tags", a.ko.Spec.Tags, b.ko.Spec.Tags)
+		}
+	}
+}
```
a-hilaly, line 184: Note to myself: need to refactor all the tagging logic like we did for ec2-controller

PR #200, opened 2024-11-25
Allow storage update operation to unavailable db instance

michaelhtm:
Issue [#2214](https://github.com/aws-controllers-k8s/community/issues/2214)

Description of changes:
With this change we allow the controller to make a storage update 
call when the DB instance is unavailable due to a `storage-full` status.

This case is not easy to test, since the minimum storage to allocate is 20GiB.
Options for testing:
* Create an existing DB instance that is already unavailable due to `storage-full`
  status and update storage (this would then require us to fill up the remaining space)

more ideas are welcome...


templates/hooks/db_instance/sdk_update_pre_build_request.go.tpl
```diff
@@ -14,7 +14,7 @@
 		ackcondition.SetSynced(desired, corev1.ConditionTrue, nil, nil)
 		return desired, nil
 	}
-	if !instanceAvailable(latest) {
+	if !instanceAvailable(latest) && !strings.Contains(*latest.ko.Status.DBInstanceStatus, "storage-full") && !delta.DifferentAt("Spec.AllocatedStorage") {
```
a-hilaly, line 17: nit: Can we create a helper function called `needStorageUpdate` and we do something like:
```
if !instanceAvailable(latest) && if !needStorageUpdate(latest) { ... }
```

PR #221, opened 2025-05-12
fix: ignore setting sdkUpdate output into DBInstance spec

michaelhtm:
Issue [#2086](https://github.com/aws-controllers-k8s/community/issues/2086), [#2123](https://github.com/aws-controllers-k8s/community/issues/2123)

Description of changes:
When we make a ModifyDBInstance call, the output returned is outdated. 
Currently we've been relying on `pendingModifiedValues` to return the 
desired field values, but based on the logs, pendingModifiedValues is not 
a reliable source.

These changes ignore the sdkUpdate output spec completely, and only patch the status.

One more change here is adding support of late initializing PerformanceInsightsKMSKeyID and PerformanceInsightsRetentionPeriod fields only when PerformanceInsights is enabled.


apis/v1alpha1/generator.yaml
```diff
@@ -298,6 +298,8 @@ resources:
     hooks:
       delta_pre_compare:
         template_path: hooks/db_instance/delta_pre_compare.go.tpl
+      # sdk_delete_post_request:
+      #   code: return r, fmt.Errorf("wait for delete")
```
michaelhtm, line 302: should we do something like this instead? ignoring sdkUpdate output? https://github.com/aws-controllers-k8s/efs-controller/blob/af06340fb3246dc107feac45136b465ade226d3c/generator.yaml#L59-L60

knottnt (commented):
Just left a few clarifying questions.

knottnt:
@michaelhtm: Would it be possible to create a test that reproduces the bug? Having that test would both make us a bit more confident in this fix and make sure that we don't accidentally have a regression in future changes.

pkg/resource/db_instance/sdk.go
```diff
@@ -1907,26 +1907,29 @@ func (rm *resourceManager) sdkUpdate(
...
 	}()
+	res := desired.ko.DeepCopy()
+	res.Status = latest.ko.Status
+
 	if instanceDeleting(latest) {
 		msg := "DB instance is currently being deleted"
-		ackcondition.SetSynced(desired, corev1.ConditionFalse, &msg, nil)
-		return desired, requeueWaitWhileDeleting
+		ackcondition.SetSynced(&resource{res}, corev1.ConditionFalse, &msg, nil)
+		return &resource{res}, requeueWaitWhileDeleting
```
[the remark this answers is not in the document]
michaelhtm: This ensures that the status of the latest is patched (resource state for example, whether it is creating, updating, etc.)
The output of an sdkUpdate ends up here:
https://github.com/aws-controllers-k8s/runtime/blob/dc7be0dac4ba4706eb3ac1153c6921c59089a5db/pkg/runtime/reconciler.go#L756-L768

And if latest == desired, the status that is returned from sdkFind will be discarded.

We only need to patch the status returned from AWS API, because we want Spec to be user defined.

test/e2e/tests/test_db_instance.py
```diff
@@ -408,3 +408,59 @@ def test_state_field_flapping(
             assert "PendingModifiedValues" in latest
             assert "MultiAZ" in latest["PendingModifiedValues"]
             assert bool(latest["PendingModifiedValues"]["MultiAZ"])
+
+    def test_crud_postgres14_update_multiAZ(
```
knottnt, line 412: Nice test!

knottnt (approved):

knottnt:
/lgtm
/hold

Just adding the hold in case we want a second reviewer on this one.

generator.yaml
```diff
@@ -298,6 +298,8 @@ resources:
     hooks:
       delta_pre_compare:
         template_path: hooks/db_instance/delta_pre_compare.go.tpl
+      # sdk_delete_post_request:
+      #   code: return r, fmt.Errorf("wait for delete")
```
knottnt, line 302: This commented out config should be deleted.

pkg/resource/db_instance/delta.go
```diff
@@ -63,14 +63,9 @@ func newResourceDelta(
...
-		b.ko.Spec.PerformanceInsightsRetentionPeriod != nil &&
-		*b.ko.Spec.PerformanceInsightsRetentionPeriod == ServiceDefaultInsightsRetentionPeriod {
-		a.ko.Spec.PerformanceInsightsRetentionPeriod = b.ko.Spec.PerformanceInsightsRetentionPeriod
-	}
-	if a.ko.Spec.PerformanceInsightsKMSKeyID == nil &&
-		b.ko.Spec.PerformanceInsightsKMSKeyID != nil {
-		a.ko.Spec.PerformanceInsightsKMSKeyID = b.ko.Spec.PerformanceInsightsKMSKeyID
+	if a.ko.Spec.PerformanceInsightsEnabled == nil &&
+		b.ko.Spec.PerformanceInsightsEnabled != nil {
+		a.ko.Spec.PerformanceInsightsEnabled = func() *bool { a := false; return &a }()
```
knottnt, line 68: Q: Is this mutating PerformanceInsightEnabled to false?

knottnt: If that's the case it seems odd to mutate the resource in the delta. I wouldn't expect the delta comparison to have any side effects.

templates/hooks/db_instance/sdk_update_pre_set_output.go.tpl
```diff
@@ -0,0 +1,6 @@
+	ko.Status = latest.ko.Status
+	setLastAppliedSecretReferenceAnnotation(&resource{ko})
```
knottnt, line 2: Q: Any reason this has been moved to out of the `sdk_update_post_set_output` hook?

michaelhtm: because we no longer reach `post_set_output`. We return here: https://github.com/aws-controllers-k8s/rds-controller/pull/221/files#diff-20a6e390c0105317eb581bd5f0bf409c375bbd090d162881ed810952ca8aef93R1998

pkg/resource/db_instance/delta.go
michaelhtm: We do this in a lot of controllers. The expected outcome if the field is missing is disabling it

knottnt (approved):

PR #242, opened 2025-09-19
wait for ClusterEndpoint to be Available before update

michaelhtm:


templates/hooks/db_cluster_endpoint/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,16 @@
+	if !clusterEndpointAvailable(latest) {
```
knottnt, line 1: The update operation might be valid when the endpoint is in the `inactive` state as well. 

```
The current status of the endpoint. One of: creating , available , deleting ,
inactive , modifying . The inactive state applies to an endpoint that can't be
used for a certain kind of cluster, such as a writer endpoint for a read-only
secondary cluster in a global database.
```

knottnt (approved):
/lgtm

PR #276, opened 2026-03-13
chore: update PerformanceInsights parameters and ignore additional parameters when dbcluster is enabled on dbinstance

shabbskagalwala:
Fixed terminal errors when PerformanceInsights* parameters cannot be changed when db instance is part of a dbcluster

```
  conditions:
    - message: >-
        api error InvalidParameterCombination: EnablePerformanceInsights
        conflicts with cluster level parameter.
      status: 'True'
      type: ACK.Terminal
    - lastTransitionTime: '2026-03-12T22:52:45Z'
      message: Resource not synced
      reason: resource is in terminal condition
      status: 'False'
      type: ACK.ResourceSynced
    - lastTransitionTime: '2026-03-12T22:52:45Z'
      message: >-
        api error InvalidParameterCombination: EnablePerformanceInsights
        conflicts with cluster level parameter.
 ```

Note to reviewers, the fields are not set on the dbinstance only set on the dbcluster

Issue #, if available:

Description of changes:

### DIFF
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -367,6 +367,21 @@ resources:
       # custom compare will ensure that we check
       # these fields delta only when DBClusterIdentifier
       # is not set
+      PerformanceInsightsEnabled:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
+      PerformanceInsightsKMSKeyID:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
+      PerformanceInsightsRetentionPeriod:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
       DatabaseInsightsMode:
         late_initialize:
           skip_incomplete_check: {}
@@ -447,12 +462,6 @@ resources:
         from:
           operation: CreateDBInstanceReadReplica
           path: PreSignedUrl
-      PerformanceInsightsKMSKeyID:
-        late_initialize:
-          skip_incomplete_check: {}
-      PerformanceInsightsRetentionPeriod:
-        late_initialize:
-          skip_incomplete_check: {}
       IOPS:
         late_initialize:
           skip_incomplete_check: {}
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -379,6 +379,21 @@ resources:
           is_ignored: true
       PerformanceInsightsRetentionPeriod:
         late_initialize:
+           skip_incomplete_check: {}
+        compare:
+          is_ignored: true
+      MasterUsername:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
+      DBName:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
+      EngineVersion:
+        late_initialize:
           skip_incomplete_check: {}
         compare:
           is_ignored: true
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -7,7 +7,7 @@ api_directory_checksum: a35626c9dd4783afbee4b0991d2ed63ab37f7444
 api_version: v1alpha1
 aws_sdk_go_version: v1.32.6
 generator_config_info:
-  file_checksum: 3c26c4c93cdd6cd3766986a1327343c5ba10e01c
+  file_checksum: bb2a500902f2ba85ff261893265d338943d21de1
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2026-03-12T18:06:01Z"
-  build_hash: 5ac6c79fbc941c426d8b70cba768820fc9296542
+  build_date: "2026-03-17T19:21:21Z"
+  build_hash: f0b080577c1ea030a347541b1f7a81843f33c9b6
   go_version: go1.25.7
-  version: v0.58.0
+  version: v0.58.0-2-gf0b0805
 api_directory_checksum: a35626c9dd4783afbee4b0991d2ed63ab37f7444
 api_version: v1alpha1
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by shabbskagalwala** [comment]
I think the error is different. It could probably be coming from here https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/hooks.go#L139 

Is there a reason it is explicitly set to false, instead of what is in the latest?

Can this be changed to ?

```
a.ko.Spec.PerformanceInsightsEnabled = b.ko.Spec.PerformanceInsightsEnabled
```

File: apis/v1alpha1/generator.yaml [hunk 1]
```diff
@@ -367,6 +367,21 @@ resources:
...
+      DBName:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
+      EngineVersion:
+        late_initialize:
+          skip_incomplete_check: {}
+        compare:
+          is_ignored: true
```
shabbskagalwala, line 399: Adding fields to ignore else the controller is in constant reconcile loop 

This db instance is part of a db cluster

```
{
  "level": "info",
  "ts": "2026-03-13T05:17:50.407Z",
  "logger": "ackrt",
  "msg": "desired resource state has changed",
  "kind": "DBInstance",
  "namespace": "mytest",
  "name": "temp-db-cluster-instance",
  "account": "REDACTED",
  "role": "",
  "region": "us-west-2",
  "is_adopted": true,
  "generation": 5,
  "diff": [
    {
      "Path": {
        "Parts": [
          "Spec",
          "DBName"
        ]
      },
      "A": null,
      "B": "tempdb"
    },
    {
      "Path": {
        "Parts": [
          "Spec",
          "EngineVersion"
        ]
      },
      "A": null,
      "B": "16.8"
    },
    {
      "Path": {
        "Parts": [
          "Spec",
          "MasterUsername"
        ]
      },
      "A": null,
      "B": "superadmin"
    }
  ]
}
```

knottnt:
> I think the error is different. It could probably be coming from here
> 
> https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/hooks.go#L139
> 
> Is there a reason it is explicitly set to false, instead of what is in the latest?
> 
> Can this be changed to ?
> 
> ```
> a.ko.Spec.PerformanceInsightsEnabled = b.ko.Spec.PerformanceInsightsEnabled
> ```

@michaelhtm do you recall why this is always set to false?

@shabbskagalwala Regarding the fix. If PerformanceInsightsEnabled should be managed by the DBCluster when the DBInstance's `Spec.DBClusterIdentifier` is set we could ignore it in the generated delta and add PerformanceInsightsEnabled under this [check](https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/hooks.go#L154) in custom_pre_compare. [newUpdateRequestPayload](https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/sdk.go#L1750) will then omit setting PerformanceInsightsEnabled in the ModifyDBInstance payload.

shabbskagalwala:
> > I think the error is different. It could probably be coming from here
> > https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/hooks.go#L139
> > 
> > Is there a reason it is explicitly set to false, instead of what is in the latest?
> > Can this be changed to ?
> > ```
> > a.ko.Spec.PerformanceInsightsEnabled = b.ko.Spec.PerformanceInsightsEnabled
> > ```
> 
> @michaelhtm do you recall why this is always set to false?
> 
> @shabbskagalwala Regarding the fix. If PerformanceInsightsEnabled should be managed by the DBCluster when the DBInstance's `Spec.DBClusterIdentifier` is set we could ignore it in the generated delta and add PerformanceInsightsEnabled under this [check](https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/hooks.go#L154) in custom_pre_compare. [newUpdateRequestPayload](https://github.com/aws-controllers-k8s/rds-controller/blob/d8f8a015fec74f68aa965f127c0d2a8286a41936/pkg/resource/db_instance/sdk.go#L1750) will then omit setting PerformanceInsightsEnabled in the ModifyDBInstance payload.

@knottnt updated the PR with the fixes/recommendation

apis/v1alpha1/generator.yaml
michaelhtm: I think having late initialize would be enough in these cases.. Actually I believe we may want to have a discussion of how we want to solve this problem.

DBInstance/DBCluster have an interesting relationship, where DBCluster owns some of the dbInstance fields. 
We may want to solve this using hooks only (maybe completely ignore some of the fields in sdkFind if the DBClusterIdentifier is not nil..Any thouughts?

michaelhtm:
>@michaelhtm do you recall why this is always set to false?

It's not always set to false, only when the user did not define a value. We want to default to false.

shabbskagalwala:
> > @michaelhtm do you recall why this is always set to false?
> 
> It's not always set to false, only when the user did not define a value. We want to default to false.

Ideally shouldn't it be the value returned from aws? Else this becomes an issue when adopting resources :)

apis/v1alpha1/generator.yaml
knottnt: @michaelhtm 

> I think having late initialize would be enough in these cases.

I don't think this is the case. If the DBCluster changes the value of these fields after the DBInstance has late initialized them the controller will still see a delta and attempt to update them directly on the DBInstance. 

> We may want to solve this using hooks only (maybe completely ignore some of the fields in sdkFind if the DBClusterIdentifier is not nil..Any thouughts

That could help with new resources as long as a the fields are all kept nil in the desired spec. However, we need to also consider existing DBInstance deployments which may have already set these fields. At least in the case of PerformanceInsightsKMSKeyID and PerformanceInsightsRetentionPeriod the fields were already being late initialized.

Given, the above I think we'll still want to ensure that cluster managed fields are omitted from Update API payload to avoid the controller getting stuck in an invalid state when the DBCluster changes.

knottnt (approved):
@shabbskagalwala thanks for this PR! Discussed with @michaelhtm offline and we're good to merge this.

/lgtm

PR #279, opened 2026-04-08
fix: prevent reconcile loops on dbcluster when autoMinorVersionUpgrad…

shabbskagalwala:
…e is true

Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/2850 (specifically for the dbcluster)

Description of changes:

When autoMinorVersionUpgrade is enabled on a DBCluster, AWS may automatically upgrade the engine minor version (e.g. 16.8 → 16.11). The controller detects a diff between the desired spec and the observed state on every reconcile, logging "desired resource state has changed" and causing the dbcluster resource to be in an unhealthy state.

This PR attempts to fix that by adding a version compare with autominorupgrade check in the delta_pre_compare hook 

### DIFF
--- a/pkg/resource/db_cluster/delta.go
+++ b/pkg/resource/db_cluster/delta.go
@@ -53,6 +53,21 @@ func newResourceDelta(
 		b.ko.Spec.StorageType = aws.String("aurora")
 	}
 
+	// When autoMinorVersionUpgrade is enabled and the engine version
+	// difference is only a minor version change (same major version),
+	// normalize the desired engine version to match the latest. This
+	// prevents the delta from firing on every reconcile when AWS
+	// auto-upgrades the minor version.
+	if a.ko.Spec.EngineVersion != nil && b.ko.Spec.EngineVersion != nil {
+		autoMinorVersionUpgrade := true
+		if a.ko.Spec.AutoMinorVersionUpgrade != nil {
+			autoMinorVersionUpgrade = *a.ko.Spec.AutoMinorVersionUpgrade
+		}
+		if !requireEngineVersionUpdate(a.ko.Spec.EngineVersion, b.ko.Spec.EngineVersion, autoMinorVersionUpgrade) {
+			a.ko.Spec.EngineVersion = b.ko.Spec.EngineVersion
+		}
+	}
+
 	compareSecretReferenceChanges(delta, a, b)
 
 	if ackcompare.HasNilDifference(a.ko.Spec.AllocatedStorage, b.ko.Spec.AllocatedStorage) {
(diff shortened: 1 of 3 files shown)


pkg/resource/db_cluster/delta.go
```diff
@@ -53,6 +53,21 @@ func newResourceDelta(
 		b.ko.Spec.StorageType = aws.String("aurora")
 	}
 
+	// When autoMinorVersionUpgrade is enabled and the engine version
+	// difference is only a minor version change (same major version),
+	// normalize the desired engine version to match the latest. This
+	// prevents the delta from firing on every reconcile when AWS
+	// auto-upgrades the minor version.
+	if a.ko.Spec.EngineVersion != nil && b.ko.Spec.EngineVersion != nil {
+		autoMinorVersionUpgrade := true
```
michaelhtm, line 62: Why do we want to set this to true by default?

templates/hooks/db_cluster/delta_pre_compare.go.tpl
michaelhtm, line 1: The issue seems to be, the EngineVersion is stale in the resource spec, due to autoMinorVersionUpgrade being true.
Should we leave it as stale? or should we consider setting it to the latest version? (although this would be considered changing a user set field)

michaelhtm (commented):
Thanks @shabbskagalwala! 
left a couple of questions below
/test all

templates/hooks/db_cluster/delta_pre_compare.go.tpl
shabbskagalwala: Imho it should be left stale, since this would be just minor version upgrade as part of the auto upgrade being set to true. The same behavior is on db instance when it's a part of db cluster (has db cluster identifier set). The version is set on db instance spec by the controller and later ignored if dbclusteridentifier is not nil.

pkg/resource/db_cluster/delta.go
shabbskagalwala: I read more into the docs and it seems like (from my understanding this is the default behavior - always on) - thats why setting it to `true`

> AutoMinorVersionUpgrade
Specifies whether minor engine upgrades are applied automatically to the DB cluster during the maintenance window. **By default, minor engine upgrades are applied automatically.**

> Valid for Cluster Type: Aurora DB clusters and Multi-AZ DB cluster.

> For more information about automatic minor version upgrades, see [Automatically upgrading the minor engine version](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Upgrading.html#USER_UpgradeDBInstance.Upgrading.AutoMinorVersionUpgrades).

> Type: Boolean

> Required: No

Happy to change it to whatever is the right approach :)

knottnt: Looks like this also matches our logic in [newCustomUpdateRequestPayload](https://github.com/aws-controllers-k8s/rds-controller/blob/1ba35e2d7a0ba456679d8bf85d36a6d7619f15b4/pkg/resource/db_cluster/custom_update.go#L570). Keeping this as `true` sounds good to me.

templates/hooks/db_cluster/delta_pre_compare.go.tpl
knottnt: Agree the value should be left stale here. Setting the user defined spec field could also lead to GitOps tools like ArgoCD viewing the DBCluster resource as out of sync.

knottnt (approved):
/lgtm

PR #282, opened 2026-04-27
Ignore DBInstance StorageType when owned by a DBCluster

knottnt:
Issue #, if available: [2683](https://github.com/aws-controllers-k8s/community/issues/2683#issuecomment-4235544670)

Description of changes:
When a DBInstance is owned by a DBCluster, including `Spec.StorageType` in the `ModifyDBInstance` payload results in an error similar to the below. When the DBCluster itself updates the storage type this can result in a persistent error on the underlying DBInstance(s) as the ACK Controller interprets the new value set by the DBCluster as delta and attempts to update the DBInstances storage type. This PR resolves the issue by ignoring `Spec.StorageType` in delta.go when `Spec.DBClusterIdentifier` has been set to avoid the change being detected as drift and to omit the value in any `ModifyDBInstance` payload when another value is changed.  

```
api error InvalidParameterCombination: You can''t specify the storage type for a DB instance in an Aurora DB cluster.'
```

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2026-04-22T18:28:27Z"
-  build_hash: c55e8227b1ff5ac6a8b8e1421a2f2076fa6a77bd
+  build_date: "2026-04-27T22:01:58Z"
+  build_hash: a8dd015d9d12fcefe9bacd128eed5c2308c19e4d
   go_version: go1.26.2
-  version: v0.58.1
+  version: v0.58.1-1-ga8dd015
 api_directory_checksum: a35626c9dd4783afbee4b0991d2ed63ab37f7444
 api_version: v1alpha1
 aws_sdk_go_version: v1.32.6
 generator_config_info:
-  file_checksum: c5c2439f981fa274d188720d9913aba340548118
+  file_checksum: 5a894c6aa0a2477572941c1b12912e4fa3428a09
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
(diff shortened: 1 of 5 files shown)


gustavodiaz7722:
Can we update e2e test to validate this fix? I tested this locally with your changes and it seems to be working

```

        # Update DBInstance StorageType (field managed by DBCluster).
        # When a DBInstance is owned by a DBCluster, including StorageType in
        # the ModifyDBInstance payload results in an
        # InvalidParameterCombination error. The controller should ignore
        # StorageType drift when DBClusterIdentifier is set.
        # See: https://github.com/aws-controllers-k8s/community/issues/2683
        cr = k8s.get_resource(db_instance_ref)
        original_storage_type = cr["spec"].get("storageType", None)

        updates = {
            "spec": {
                "storageType": "aurora"
            }
        }
        k8s.patch_custom_resource(db_instance_ref, updates)
        time.sleep(CHECK_WAIT_AFTER_REF_RESOLVE_SECONDS)

        # The resource should remain synced because StorageType changes are
        # ignored for cluster-owned instances (no ModifyDBInstance call).
        assert k8s.wait_on_condition(db_instance_ref, "ACK.ResourceSynced", "True", wait_periods=CHECK_WAIT_AFTER_REF_RESOLVE_SECONDS)

        # Verify no terminal condition was set (i.e. no
        # InvalidParameterCombination error from AWS).
        cr = k8s.get_resource(db_instance_ref)
        terminal_cond = k8s.get_resource_condition(db_instance_ref, "ACK.Terminal")
        assert terminal_cond is None or terminal_cond.get("status", "False") == "False"

        # Verify the AWS-side StorageType was NOT changed (the controller
        # should have ignored the delta).
        latest_aws = db_instance.get(db_instance_id)
        assert latest_aws is not None
        assert latest_aws["DBInstanceStatus"] == "available"
```

PR #283, opened 2026-04-28
fix: add AllocatedStorage to customPreCompare delta check

michaelhtm:
Issue [#2845](https://github.com/aws-controllers-k8s/community/issues/2845)

Description of changes:
AllocatedStorage was marked compare.is_ignored in generator.yaml but
was never added to the customPreCompare hook, so changes to
allocatedStorage produced an empty delta and no ModifyDBInstance call.

### DIFF
--- a/test/e2e/tests/test_db_instance.py
+++ b/test/e2e/tests/test_db_instance.py
@@ -507,3 +507,33 @@ class TestDBInstance:
         assert latest is not None
         assert latest['DBInstanceStatus'] == 'available'
         assert latest['MultiAZ'] is True
+
+    def test_update_allocated_storage(
+            self,
+            postgres14_t3_micro_instance,
+    ):
+        (ref, cr, _) = postgres14_t3_micro_instance
+        db_instance_id = cr["spec"]["dbInstanceIdentifier"]
+
+        # Wait for the resource to get synced
+        assert k8s.wait_on_condition(ref, "ACK.ResourceSynced", "True", wait_periods=MAX_WAIT_FOR_SYNCED_MINUTES)
+
+        latest = db_instance.get(db_instance_id)
+        assert latest is not None
+        assert latest['DBInstanceStatus'] == 'available'
+        assert latest['AllocatedStorage'] == 5
+
+        updates = {
+            "spec": {"allocatedStorage": 10},
+        }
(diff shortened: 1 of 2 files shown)


test/e2e/tests/test_db_instance.py
```diff
@@ -507,3 +507,31 @@ def test_crud_postgres14_update_multiAZ(
...
+        assert latest['DBInstanceStatus'] == 'available'
+        assert latest['AllocatedStorage'] == 5
+
+        updates = {
+            "spec": {"allocatedStorage": 10},
+        }
+        k8s.patch_custom_resource(ref, updates)
+        time.sleep(MODIFY_WAIT_AFTER_SECONDS)
+
+        assert k8s.wait_on_condition(ref, "ACK.ResourceSynced", "True", wait_periods=MAX_WAIT_FOR_SYNCED_MINUTES)
```
knottnt: We may need to wait for awhile longer if we want this test. It looks like RDS instances can be in the storage-optimization state for quite a while. That said if it really does take hours we may want to mark this as `run_slow`.

> Typically, storage optimization completes in a few hours, but the process can take more than 24 hours. The instance is operational during storage optimization and your application is still available. You can run your applications during storage optimization.


https://repost.aws/knowledge-center/rds-stuck-in-storage-optimization

knottnt (commented):
@michaelhtm change looks good, but the new test appears to be failing due to the resource not returning to the `available` status for an extended period of time.

test/e2e/tests/test_db_instance.py
```diff
@@ -507,3 +507,29 @@ def test_crud_postgres14_update_multiAZ(
...
+        updates = {
+            "spec": {"allocatedStorage": 10},
+        }
+        k8s.patch_custom_resource(ref, updates)
+        time.sleep(35)
+
+        condition.assert_not_synced(ref)
+        cr = k8s.get_resource(ref)
+        assert cr is not None
+        assert cr['spec']['allocatedStorage'] == 10
```
knottnt, line 535: This doesn't quite validate that the change was successfully made to the AWS resource. In addition to checking the k8s resource would you be able to check that the AWS resource's status has changed to the `storage-optimization` state and also reports the new allocatedStorage value?

knottnt (approved):
/lgtm

PR #287, opened 2026-05-12
Fix DBInstance AWS defaults causing permanent diff after adoption via adopt-or-create

knottnt:
Issue #, if available: [2617](https://github.com/aws-controllers-k8s/community/issues/2617)

Description of changes:
Adds MonitoringInterval, PubliclyAccessible, and CopyTagsToSnapshot to the late initialization configuration for DBInstance. Without this, adopting an existing DBInstance with `adopt-or-create` without explicitly setting the field results in a permanent diff. The controller consistently sees these empty fields as drifted from the defaults set by AWS. With the PR the controller will pick up the AWS default values after adoption when unset preventing subsequent reconciliation loops from incorrectly seeing an erroneous delta. 


michaelhtm (approved):
/lgtm

PR #293, opened 2026-05-26
feat: add DBInstanceParameterGroupName field to DBCluster

### LINKED ISSUE #2778: ACK RDS Controller: Support DBInstanceParameterGroupName in DBCluster for Aurora major version upgrade
**Describe the bug**

When performing a **major engine version upgrade** on **aurora-postgresql RDS cluster** that uses a custom db instance parameter group. The upgrade fails with the following error:

`api error InvalidParameterCombination: The current DB instance parameter group <custom_parameter_group_name> is custom. You must explicitly specify a new DB instance parameter group, either default or custom, for the engine version upgrade.`

Currently, the **DBCluster CRD** supports only `dbClusterParameterGroupName`. It does not support the `dbInstanceParameterGroupName`, which is available in the underlying AWS `ModifyDBCluster` API. Because of this limitation, the controller cannot pass the required instance level parameter group during a major version upgrade.

**Steps to reproduce**
1. Create Aurora PostgreSQL db cluster using the default cluster parameter group
2. Create a custom db instance parameter group
3. Associate the custom instance parameter group with the db instance
4. Upgrade the cluster to new major engine version e.g 16.x to 17.x

**Expected outcome**
The cluster and it's instances are successfully upgraded to the new major engine version using the specified custom parameter group, with no errors. The updates should be visible in aws console.

**Environment**

* Kubernetes version: 1.34
* AWS service targeted: RDS (Aurora PostgreSQL)
* Current RDS controller version: `1.7.4`

michaelhtm:
Description of changes:
The field is sourced from ModifyDBCluster and passed to AWS only during
major engine version upgrades, enabling Aurora clusters with custom
instance parameter groups to upgrade successfully.

Mark field with compare.is_ignored to prevent infinite reconciliation
since AWS never returns it in DescribeDBClusters. 

Resolves aws-controllers-k8s/community#2778

### DIFF
--- a/pkg/resource/db_cluster/custom_update.go
+++ b/pkg/resource/db_cluster/custom_update.go
@@ -574,6 +574,14 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 		}
 		if requireEngineVersionUpdate(desired.ko.Spec.EngineVersion, latest.ko.Spec.EngineVersion, autoMinorVersionUpgrade) {
 			res.EngineVersion = desired.ko.Spec.EngineVersion
+			// AWS only accepts DBInstanceParameterGroupName in combination
+			// with AllowMajorVersionUpgrade, and only for a major version
+			// upgrade. Sending it on a minor version change fails with
+			// InvalidParameterCombination.
+			if desired.ko.Spec.DBInstanceParameterGroupName != nil &&
+				isMajorVersionUpgrade(desired.ko.Spec.EngineVersion, latest.ko.Spec.EngineVersion) {
+				res.DBInstanceParameterGroupName = desired.ko.Spec.DBInstanceParameterGroupName
+			}
 		}
 	}
 	if desired.ko.Spec.MasterUserPassword != nil && delta.DifferentAt("Spec.MasterUserPassword") {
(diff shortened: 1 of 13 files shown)


pkg/resource/db_cluster/custom_update.go
```diff
@@ -573,6 +573,9 @@ func (rm *resourceManager) newCustomUpdateRequestPayload(
 		}
 		if requireEngineVersionUpdate(desired.ko.Spec.EngineVersion, latest.ko.Spec.EngineVersion, autoMinorVersionUpgrade) {
 			res.EngineVersion = desired.ko.Spec.EngineVersion
+			if desired.ko.Spec.DBInstanceParameterGroupName != nil {
```
knottnt: Q: From the API docs it sounds like this field is only valid during a major version upgrade. Do we need to check that `DBInstanceParameterGroupName` is in the delta before including it to prevent an error during a subsequent minor version upgrade?

pkg/resource/db_cluster/custom_update_test.go
```diff
@@ -104,3 +104,38 @@ func TestNewCustomUpdateRequestPayload_PreferredBackupWindowNotInDelta(t *testin
 	assert.NotNil(t, input)
 	assert.Nil(t, input.PreferredBackupWindow)
 }
+
+func TestNewCustomUpdateRequestPayload_DBInstanceParameterGroupName_MajorUpgrade(t *testing.T) {
```
knottnt, line 108: Do we need a test for the below scenarios?
1. No engine version change
2. A minor version engine change.

test/e2e/tests/test_db_cluster.py
knottnt, line 1: Would be good to have an e2e test for the major version engine upgrade case.

PR #295, opened 2026-05-28
fix: prevent AllocatedStorage writeback to DBCluster spec

### LINKED ISSUE #2434: ACK-RDS DBCluster adds invalid AllocatedStorage to aurora cluster
**Describe the bug**
The `DBCluster` resource always adds a `spec.allocatedStorage: 1` when none is specified. The controller then fails to create `aurora-postgresql` type clusters since the `allocatedStorage` field isn't supported.

**Steps to reproduce**

Create a simple `DBCluster`:
```
kind: DBCluster
metadata:
  name: test-db-instance
spec:
  autoMinorVersionUpgrade: true
  databaseName: test_db
  dbClusterIdentifier: test-db-instance
  dbClusterParameterGroupName: default.aurora-postgresql16
  dbSubnetGroupName: test-db-instance
  deletionProtection: true
  enableCloudwatchLogsExports: []
  enableIAMDatabaseAuthentication: false
  enablePerformanceInsights: false
  engine: aurora-postgresql
  engineVersion: '16.6'
  masterUserPassword:
    key: DB_PASSWORD
    name: test-db-instance
  masterUsername: postgres
  vpcSecurityGroupIDs:
    - sg-CHANGE_ME
```

**Expected outcome**
- The ACK-RDS controller shouldn't mutate the DBCluster resource's `spec.allocatedStorage` field.
- The DBCluster should be created.

**Actual outcome**

Here's the live `DBCluster` resource (I redacted fields that weren't relevant). 
We can see that `spec.allocatedStorage`  was set to `1` and that the DBCluster creation failed since that field was included in the payload sent to AWS.
```
kind: DBCluster
metadata:
  name: test-db-instance
spec:
  allocatedStorage: 1
[...]
status:
[...]
  conditions:
    - message: "InvalidParameterCombination: AllocatedStorage isn't supported for DB engine aurora-postgresql.\n\tstatus code: 400, request id: [...]"
      status: 'True'
      type: ACK.Terminal
[...]
```

**Environment**

* Kubernetes version: `1.31`
* Using EKS (yes/no), if so version?: `eks.21`
* AWS service targeted (S3, RDS, etc.): `RDS`

michaelhtm:
Description of changes:
AWS always returns AllocatedStorage=1 for Aurora clusters even though
it's not a user-configurable value for Aurora engines. The generated
code was writing this value back into the Kubernetes resource's Spec,
causing subsequent create retries to fail with
"AllocatedStorage isn't supported for DB engine aurora-postgresql".

Resolves aws-controllers-k8s/community#2434

### DIFF
--- a/pkg/resource/db_cluster/hooks.go
+++ b/pkg/resource/db_cluster/hooks.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	svcapitypes "github.com/aws-controllers-k8s/rds-controller/apis/v1alpha1"
 	"github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1"
@@ -151,6 +152,23 @@ func clusterDeleting(r *resource) bool {
 	}
 	dbcs := *r.ko.Status.Status
 	return dbcs == StatusDeleting
+}
+
+// isAuroraEngine returns true if the engine string indicates an Aurora cluster
+// (aurora, aurora-mysql, aurora-postgresql).
+func isAuroraEngine(engine *string) bool {
+	if engine == nil {
+		return false
+	}
+	return strings.HasPrefix(*engine, "aurora")
+}
+
+// clearAuroraAllocatedStorage nils out AllocatedStorage for Aurora engines.
+// AWS always returns AllocatedStorage=1 for Aurora but rejects it on input.
+func clearAuroraAllocatedStorage(ko *svcapitypes.DBCluster) {
+	if isAuroraEngine(ko.Spec.Engine) {
+		ko.Spec.AllocatedStorage = nil
+	}
 }
 
 // syncTags keeps the resource's tags in sync
@@ -299,6 +317,7 @@ func (rm *resourceManager) restoreDbClusterFromSnapshot(
 
 	rm.setResourceFromRestoreDBClusterFromSnapshotOutput(r, resp)
 	rm.setStatusDefaults(r.ko)
+	clearAuroraAllocatedStorage(r.ko)
 
 	// We expect the DB cluster to be in 'creating' status since we just
 	// issued the call to create it, but I suppose it doesn't hurt to check
@@ -332,6 +351,7 @@ func (rm *resourceManager) restoreDbClusterToPointInTime(
 
 	rm.setResourceFromRestoreDBClusterToPointInTimeOutput(r, resp)
 	rm.setStatusDefaults(r.ko)
+	clearAuroraAllocatedStorage(r.ko)
 
 	// We expect the DB cluster to be in 'creating' status since we just
 	// issued the call to create it, but I suppose it doesn't hurt to check
--- a/templates/hooks/db_cluster/sdk_read_many_post_set_output.go.tpl
+++ b/templates/hooks/db_cluster/sdk_read_many_post_set_output.go.tpl
@@ -13,6 +13,7 @@
 	} else {
 		ackcondition.SetSynced(&resource{ko}, corev1.ConditionTrue, nil, nil)
 	}
+	clearAuroraAllocatedStorage(ko)
 	if len(r.ko.Spec.VPCSecurityGroupIDs) > 0 {
 		// If the desired resource has security groups specified then update the spec of the latest resource with the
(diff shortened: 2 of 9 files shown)


templates/hooks/db_cluster/sdk_read_many_post_set_output.go.tpl
```diff
@@ -13,6 +13,15 @@
 	} else {
 		ackcondition.SetSynced(&resource{ko}, corev1.ConditionTrue, nil, nil)
 	}
+	// AllocatedStorage is only valid for Multi-AZ DB clusters (where it's
+	// required and always user-specified). For Aurora clusters, AWS always
+	// returns AllocatedStorage=1 which is meaningless. Without this guard,
+	// that value gets written back to the Spec and sent on subsequent API
+	// calls, causing "AllocatedStorage isn't supported for DB engine
+	// aurora-*" errors.
+	if r.ko.Spec.AllocatedStorage == nil {
```
knottnt: Q: Is this correct behavior when adopting a DBCluster with the `adopt` policy? I'm not sure we want to omit it in that case.

pkg/resource/db_cluster/hooks.go
```diff
@@ -297,8 +297,18 @@ func (rm *resourceManager) restoreDbClusterFromSnapshot(
 		return nil, respErr
 	}
 
+	// AllocatedStorage is only valid for Multi-AZ DB clusters (where it's
+	// required and always user-specified). For Aurora clusters, AWS always
```
knottnt: Q: Is it the case that AllocatedStorage is always user specified in the spec when restoring from a snapshot? Looking at the [RestoreDBClusterFromSnapshot API Reference](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBClusterFromSnapshot.html) AllocatedStorage isn't in the input shape, but is returned in the output.

templates/hooks/db_cluster/sdk_read_many_post_set_output.go.tpl
michaelhtm: you're right. Updated to only ignore on aurora changes

pkg/resource/db_cluster/hooks.go
```diff
@@ -153,6 +154,16 @@ func clusterDeleting(r *resource) bool {
 	return dbcs == StatusDeleting
 }
 
+// isAuroraEngine returns true if the engine string indicates an Aurora cluster
+// (aurora, aurora-mysql, aurora-postgresql). AllocatedStorage is not supported
+// for Aurora engines — AWS always returns 1 but rejects it on input.
+func isAuroraEngine(engine *string) bool {
```
knottnt, line 159: nit: Could you update the delta.go hook template to use this function as well.

https://github.com/aws-controllers-k8s/rds-controller/blob/31cf74b2dca815169e6835e7b5803bb0c9df64a6/pkg/resource/db_cluster/delta.go#L50

pkg/resource/db_cluster/hooks.go
```diff
@@ -153,6 +154,16 @@ func clusterDeleting(r *resource) bool {
 	return dbcs == StatusDeleting
 }
 
+// isAuroraEngine returns true if the engine string indicates an Aurora cluster
+// (aurora, aurora-mysql, aurora-postgresql). AllocatedStorage is not supported
+// for Aurora engines — AWS always returns 1 but rejects it on input.
```
knottnt: nit: This function doesn't directly handle any AllocatedStorage logic and can be used for other purposes so its a bit confusing to have it mentioned here.
```suggestion
// isAuroraEngine returns true if the engine string indicates an Aurora cluster
// (aurora, aurora-mysql, aurora-postgresql).
```

knottnt: It would be good to document why we are clearing AllocatedStorage in the new hook logic. If the desire is to have a single place to document the behavior we could add a `clearAuroraAllocatedStorage` hook function that is called in the hooks.

knottnt (commented):
Looks good @michaelhtm. Just had a couple nits.

knottnt (approved):
Thanks @michaelhtm!

/lgtm

PR #298, opened 2026-06-05
fix: requeue after syncing parameter overrides to refresh status

michaelhtm:
Description of changes:
customUpdate was returning desired directly after modifying parameters
in AWS. Since no subsequent ReadOne occurs after a successful update,
ParameterOverrideStatuses never reflected newly added parameters. Now
we return a requeue error so the reconciler re-runs ReadOne and picks
up the correct status from AWS.

Fixes both DBClusterParameterGroup and DBParameterGroup resources.


knottnt (approved):
/lgtm

knottnt (approved):
/lgtm

PR #300, opened 2026-06-12
fix: prevent DBCluster Aurora AllocatedStorage reconcile loop

### LINKED ISSUE #2914: RDS controller is flagging a difference on allocatedStorage
**Describe the bug**
Controller is generating logs as follows:

{"level":"info","ts":"2026-06-12T04:12:34.928Z","logger":"ackrt","msg":"updated 
resource","kind":"DBCluster","namespace":"terms","name":"apg17-dbcluster","account":"000000000000","role":"","region":"eu-west-2","is_adopted":false,"generation":20}

The cluster is running Aurora Serverless v2 (17.9) and does not require this field to be specified. Even after specifying a value (1) for this field in an attempt to get the reconciliation logic to cure the condition it is still reporting the DBCluster resource as not ready. 

**Expected outcome**

Ideally specifying this should not be required for Aurora Serverless but if it is required then the controller logic should at least handle the value supplied and mark the resource as ready.

**Environment**

EKS 1.34 (bottlerocket)
RDS controller 1.9.2

michaelhtm:
Description of changes:
AllocatedStorage does not apply to Aurora clusters, but a value left in
the spec compared against the cleared observed side produced a delta on
every reconcile that ModifyDBCluster could not resolve. Clear it on the
desired side too in delta_pre_compare so Aurora clusters reach synced.

Resolves aws-controllers-k8s/community#2914

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,8 +1,8 @@
 ack_generate_info:
-  build_date: "2026-06-05T17:05:21Z"
-  build_hash: 8d8f5f2406b145d2bf19f9bba75086b2445ffe4b
+  build_date: "2026-06-12T17:01:11Z"
+  build_hash: 2970ca9b3789515150e27ab80630175a8c77cba4
   go_version: go1.26.3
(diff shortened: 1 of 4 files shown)


knottnt (approved):
/lgtm

PR #308, opened 2026-07-13
fix: prevent reconcile loop on DBInstance - send AllowMajorVersionUpgrade only on engine version change

### LINKED ISSUE #2956: rds-controller DBInstance: reconcile loop - controller always sends allowMajorVersionUpgrade=true in ModifyDBInstance
## Description

The ACK RDS controller sends `allowMajorVersionUpgrade: true` in every `ModifyDBInstance` call for `DBInstance` resources, regardless of desired state. This causes a continuous no-op reconcile loop firing every ~30 seconds.

## Steps to Reproduce

1. Create any `DBInstance` resource managed by the ACK RDS controller
2. Observe CloudTrail — `ModifyDBInstance` is called every ~30 seconds with only:
   - `applyImmediately: true`
   - `allowMajorVersionUpgrade: true`

## Evidence

~200 `ModifyDBInstance` API calls per 30 minutes across 4 DBInstance resources

## Related

This is the `DBInstance` equivalent of the `DBCluster` fix in aws-controllers-k8s/rds-controller#279 (`fix: prevent reconcile loops on dbcluster when autoMinorVersionUpgrade`). The same issue exists in the `customUpdateDBInstance` hook — `allowMajorVersionUpgrade` is not a spec field in the CRD (field not declared in schema) but the controller sends it unconditionally on every reconcile.

## Environment

- ACK RDS controller: `1.10.1`
- ack-chart: `46.129.10`
- Kubernetes: `1.36`
- Region: `us-east-1`

kappa8219:
## Description

The `AllowMajorVersionUpgrade` field was set unconditionally to `true` in every `ModifyDBInstance` call via `override_values` in `generator.yaml`. This caused a continuous reconcile loop where `ModifyDBInstance` was called every ~30 seconds even when no fields had changed.

**Root cause:** `generator.yaml` has `AllowMajorVersionUpgrade: aws.Bool(true)` as an `override_values` for `ModifyDBInstance`, which unconditionally injects this field into every update request regardless of whether the engine version is actually changing.

**Fix:** Only set `AllowMajorVersionUpgrade = true` when `Spec.EngineVersion` delta is present, matching the original intent (allow major version upgrades when the user explicitly changes the engine version).

## Evidence

~200 no-op `ModifyDBInstance` API calls per 30 minutes across 4 DBInstance resources, all containing only:
- `applyImmediately: true`
- `allowMajorVersionUpgrade: true`

## Testing

- Existing unit tests pass (`go build ./pkg/resource/db_instance/...` compiles clean)
- The fix follows the same pattern as the `ModifyDBCluster` fix in #279

## Related

- Fixes https://github.com/aws-controllers-k8s/community/issues/2956
- DBInstance equivalent of the DBCluster fix in #279

/kind bug

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -132,11 +132,13 @@ operations:
       # makes to a resource's Spec to be reconciled by the ACK service
       # controller, not a different service.
       ApplyImmediately: aws.Bool(true)
-      # We override the value of the ApplyImmediately field in the modify
-      # operations to "true" because we want changes that a Kubernetes user
-      # makes to a resource's Spec to be reconciled by the ACK service
-      # controller, not a different service.
-      AllowMajorVersionUpgrade: aws.Bool(true)
+      # NOTE: AllowMajorVersionUpgrade is intentionally NOT set here as an
+      # override_value. Setting it unconditionally caused a reconcile loop
+      # where ModifyDBInstance was called on every reconcile even when no
+      # fields had changed. Instead, AllowMajorVersionUpgrade is set
+      # conditionally in the sdk_update_post_build_request hook only when
+      # an engine version change is actually required.
+      # See: https://github.com/aws-controllers-k8s/community/issues/2956
 resources:
   DBCluster:
     update_operation:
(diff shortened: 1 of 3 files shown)


kappa8219:
@michaelhtm please take a look

michaelhtm:
Looks like issue was closed
/close

PR #315, opened 2026-08-13
feat(rds): Add twelve cross-resource references on top-level fields

gustavodiaz7722:
Adds cross-resource references to 8 top-level fields on `DBCluster`, `DBClusterEndpoint`, `DBInstance`, `DBProxy` and `GlobalCluster` in `rds-controller`.

Every field here is a **top-level** spec field (not nested in a struct or list) whose target Kind is confirmed ACK-managed, and **every one was verified on a live cluster**. None needs `set: ignore`, `late_initialize`, or `skip_resource_state_validations`.

### Fields Added

| # | Field (`generator.yaml` path) | Target | `path` |
| --- | --- | --- | --- |
| 1 | `DBCluster.SourceDBClusterIdentifier` | `DBCluster` (same-service) | `Spec.DBClusterIdentifier` |
| 2 | `DBInstance.DBClusterIdentifier` | `DBCluster` (same-service) | `Spec.DBClusterIdentifier` |
| 3 | `DBInstance.SourceDBInstanceIdentifier` | `DBInstance` (same-service) | `Spec.DBInstanceIdentifier` |
| 4 | `DBInstance.DBSnapshotIdentifier` | `DBSnapshot` (same-service) | `Spec.DBSnapshotIdentifier` |
| 5 | `DBClusterEndpoint.StaticMembers` | `DBInstance` (same-service) | `Spec.DBInstanceIdentifier` |
| 6 | `DBProxy.VpcSecurityGroupIds` | ec2 `SecurityGroup` | `Status.ID` |
| 7 | `DBProxy.VpcSubnetIds` | ec2 `Subnet` | `Status.SubnetID` |
| 8 | `GlobalCluster.SourceDBClusterIdentifier` | `DBCluster` (same-service) | `Status.ACKResourceMetadata.ARN` |

### Notes

- `DBInstance.dbClusterIdentifier` is the most user-visible of these: it is the parent link for every Aurora instance, so creating a cluster and its instances in one manifest no longer needs a hardcoded identifier.
- Three of these fields carry a pre-existing `from:` block because they are sourced from a non-Create operation (`RestoreDBClusterToPointInTime`, `CreateDBInstanceReadReplica`, `RestoreDBInstanceFromDBSnapshot`). The `references` block coexists with `from:`, and all three resolved and synced on a live cluster — this had been an open question about the generator and is now settled empirically.
- **Deliberate partial coverage.** The source and snapshot identifier fields also accept an ARN for cross-region or shared resources, which a reference cannot serve since the target CR is not in this cluster. The concrete field stays usable for those cases.
- `GlobalCluster.sourceDBClusterIdentifier` is the reverse edge of `DBCluster.globalClusterIdentifier`, which is left **unconfigured deliberately** so the pair cannot deadlock. Whichever direction is wired second owns that decision.

### Verification

Deployed this branch to a dev EKS cluster (`us-west-2`) with `ack-workspace deploy rds --resync-period 60` and exercised every reference against real AWS resources. The controller image under test was built from this branch.

This is the **lightweight** check: the reference resolves and the resource reaches `Synced`. It is deliberately not the full pass matrix — no multi-resync delta count and no reference-preservation check.

| # | Reference | RefsResolved | Synced | Concrete absent | `*Ref` present | Result |
| --- | --- | --- | --- | --- | --- | --- |
| 1 | `DBCluster.sourceDBClusterIdentifierRef` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 2 | `DBInstance.dbClusterIdentifierRef` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 3 | `DBInstance.sourceDBInstanceIdentifierRef` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 4 | `DBInstance.dbSnapshotIdentifierRef` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 5 | `DBClusterEndpoint.staticMemberRefs` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 6 | `DBProxy.vpcSecurityGroupRefs` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 7 | `DBProxy.vpcSubnetRefs` | ✅ | ✅ | ✅ | ✅ | **PASS** |
| 8 | `GlobalCluster.sourceDBClusterIdentifierRef` | ✅ | ✅ | ✅ | ✅ | **PASS** |

**Summary: 8 PASS of 8.**

- `DBCluster.sourceDBClusterIdentifierRef` — point-in-time restore into a new cluster from the referenced source
- `DBInstance.dbClusterIdentifierRef` — Aurora instance created into the referenced cluster
- `DBInstance.sourceDBInstanceIdentifierRef` — read replica created from the referenced instance
- `DBInstance.dbSnapshotIdentifierRef` — instance restored from the referenced snapshot
- `DBClusterEndpoint.staticMemberRefs` — endpoint created with the referenced instance as a static member
- `DBProxy.vpcSecurityGroupRefs` — resolved to the ACK-managed security group id
- `DBProxy.vpcSubnetRefs` — resolved to both ACK-managed subnet ids
- `GlobalCluster.sourceDBClusterIdentifierRef` — global cluster created with the referenced cluster as its member

AWS-side confirmation that each resolved value actually reached the API, rather than just that resolution reported success:

```
$ aws rds describe-db-clusters --db-cluster-identifier reftest-aurora-pitr
{"Id": "reftest-aurora-pitr", "Status": "available"}       # restored from reftest-aurora

$ aws rds describe-global-clusters --global-cluster-identifier reftest-globalcluster
{"Status": "available",
 "Members": ["arn:aws:rds:us-west-2:...:cluster:reftest-aurora-pitr"]}   # the resolved ARN

$ aws rds describe-db-instances --db-instance-identifier reftest-mysql-replica
{"Id": "reftest-mysql-replica", "Source": "reftest-mysql", "Status": "available"}
```

The `Members` and `Source` values are the resolved reference targets; neither was ever supplied literally.

### Scope: four references were removed from this PR

An earlier revision of this PR carried 12 references. The four that on-cluster testing could not verify have been **removed** rather than shipped unverified, so what remains is exactly what was proven to work:

| Removed field | Why |
| --- | --- |
| `DBCluster.DomainIAMRoleName` | Needs an AWS Directory Service directory, and the paired `domain` field has no ACK controller behind it, so the reference could only ever cover half the pairing. |
| `DBInstance.DomainIAMRoleName` | Same Directory Service dependency. |
| `DBInstance.CustomIAMInstanceProfile` | RDS Custom only, which needs a custom engine version built from a customer-supplied AMI. |
| `DBInstance.DBClusterSnapshotIdentifier` | The reference resolved correctly, but the create failed terminally with `InvalidParameterValue: Invalid master password` — see below. |

On that last one: the reference itself was fine (`ACK.ReferencesResolved=True`, concrete field absent, `*Ref` present, against a synced ACK-managed `DBClusterSnapshot`). The failure looks like the controller not routing to a restore-from-cluster-snapshot call and instead attempting a plain `CreateDBInstance`, which requires a master password. The sibling `DBSnapshotIdentifier` on the same resource *does* route correctly and is kept in this PR, so the gap is specific to the cluster-snapshot path and exists independently of this change — worth a separate look by the controller owners. The AWS model also restricts that field to Multi-AZ DB cluster snapshots ("Can't be the identifier of an Aurora DB cluster snapshot").

The pre-existing `from:` block on `DBClusterSnapshotIdentifier` is retained — only its `references` block was removed, so that field behaves exactly as it did before this PR.

### What was checked mechanically

- Workspace refreshed (runtime, code-generator, controller) before generating; regenerated with `ack-workspace build rds` against a clean code-generator at the `v0.62.1` tag, matching the stamp on `main`.
- A `*Ref` companion is emitted on the spec for all 8 fields, and the companions for the 4 removed fields are gone.
- `resolveReferenceFor<Field>` is emitted for all 8 and absent for the 4 removed, confirmed by diffing the *set* of resolver names against `main` (regeneration reorders the file, so a naive line diff over-reports).
- `vpcSubnetIDs` relaxed out of the `DBProxy` CRD `required` list.
- `service_name` omitted for same-service targets, set for cross-service.
- `go build ./cmd/controller` passes.
- No new `go.mod` dependency (iam and ec2 were already present), so `ATTRIBUTION.md` is unchanged. The `instanceprofiles` RBAC rule that the removed `InstanceProfile` reference required is dropped again.

### Commit layout

`main` carries the code-generator `v0.62.1` regeneration (from the auto-generated release PR), and this branch is merged up to date with it, so the **Files changed** diff is limited to the reference change itself plus the recorded ack-generate stamp.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,8 +1,8 @@
 ack_generate_info:
-  build_date: "2026-08-04T17:30:04Z"
-  build_hash: db232581a560896c2dc461a244f96d5bf3191ec6
-  go_version: go1.26.5
-  version: v0.62.0
+  build_date: "2026-08-13T16:48:13Z"
+  build_hash: 65d45b2e6c9efd6aca20e0d36826d1e18e4ba2b7
+  go_version: go1.26.0
+  version: v0.62.1
 api_directory_checksum: 81211ae123aafed5b831e83e8e420bb188895d79
 api_version: v1alpha1
 aws_sdk_go_version: v1.34.0
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,8 +1,8 @@
 ack_generate_info:
-  build_date: "2026-08-13T22:05:10Z"
-  build_hash: 3dfc7a1e4135d59e6b1ecf6955ab1df4e829c9b5
+  build_date: "2026-08-13T22:22:49Z"
+  build_hash: 65d45b2e6c9efd6aca20e0d36826d1e18e4ba2b7
   go_version: go1.26.0
-  version: v0.62.1-2-g3dfc7a1-dirty
+  version: v0.62.1
 api_directory_checksum: d99401cf512d4785088e63c42c4039efe8f5c2fe
 api_version: v1alpha1
 aws_sdk_go_version: v1.34.0
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2026-08-13T16:48:13Z"
+  build_date: "2026-08-13T16:53:34Z"
   build_hash: 65d45b2e6c9efd6aca20e0d36826d1e18e4ba2b7
   go_version: go1.26.0
   version: v0.62.1
-api_directory_checksum: 81211ae123aafed5b831e83e8e420bb188895d79
+api_directory_checksum: 2c13c1b02a7fadc5c9082bad1d8153238c05c117
 api_version: v1alpha1
 aws_sdk_go_version: v1.34.0
 generator_config_info:
-  file_checksum: 8bc796cb417e5f245e8f34c171853f1583e847b0
+  file_checksum: 77ed7e6d7aa712c72a8871251ad3c46e2c1feee2
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,9 +1,9 @@
 ack_generate_info:
-  build_date: "2026-08-25T20:58:21Z"
-  build_hash: 65d45b2e6c9efd6aca20e0d36826d1e18e4ba2b7
+  build_date: "2026-08-25T21:02:24Z"
+  build_hash: d4aaca5e1dd6acfd81094936dc825ad30d2bbb39
   go_version: go1.26.0
-  version: v0.62.1
-api_directory_checksum: dcd3e71094e55a7524cb2422f1bf69bbd3e03d49
+  version: v0.62.1-4-gd4aaca5
+api_directory_checksum: 2aad9b61dd0449c9d4d098ddc4acf735bbbb5662
 api_version: v1alpha1
 aws_sdk_go_version: v1.34.0
 generator_config_info:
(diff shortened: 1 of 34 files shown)


sapphirew (approved):
/lgtm

